PR #18, opened 2023-06-17
Feat/filters dropdown

AniaPiWo:


wawrzonnn (approved):

src/components/FilterButton/FilterButton.tsx
```diff
@@ -1,20 +1,93 @@
...
+            <div className={styles.buttonTextChecked}>{text}</div>
+          )}
+          {selectedOptions.length != 0 && (
+            <div className={styles.selectedCount}>{selectedOptions.length}</div>
+          )}
+          {arrow === 'IoIosArrowUp' ? <IoIosArrowUp /> : <IoIosArrowDown />}
+        </button>
+        {isDropdownActive && (
+          <div className={styles.dropdown}>
+            <form className={styles.dropdownForm}>
```
Sumick, line 70: **question:** Why `form` here?

Sumick (changes requested):
Overall looks good! I think we are missing one thing here - two very specific props, that are currently missing in this `FilterButton` component. In your current implementation, you won't be able to set value in that `FilterButton` in other place than this component body. You should be able to set all these selected values in one place - which is probably `Filters` component.

src/components/Filters/Filters.tsx
```diff
@@ -5,6 +5,60 @@ import 'material-icons/iconfont/material-icons.css'
...
+        updatedOptions = selectedSeniority.includes(option)
+          ? selectedSeniority.filter(
+              (selectedOption) => selectedOption !== option,
+            )
+          : [...selectedSeniority, option]
+        setSelectedSeniority(updatedOptions)
+        console.log(selectedSeniority)
+        break
+
+      case 'Availability':
```
Sumick, line 37: **nitpick:** You could keep these values as enum, it's quite important, and we will avoid string duplication / maintanance:

```ts
export enum FilterNames {
Availability = "Availability",
...
}

File: src/components/Filters/Filters.tsx [hunk 3]
```diff
@@ -14,17 +68,38 @@ const Filters = () => {
...
     <div className={styles.mainContainer}>
       <div className={styles.features}>
-        <FilterButton text="Technology" />
-        <FilterButton text="Seniority" />
-        <FilterButton text="Availability" />
-        <FilterButton text="Location" />
+        <FilterButton
+          text="Technology"
+          options={technologyList}
+          onSelect={(option) => handleSelect(option, 'Technology')}
```
Sumick, line 82: **nitpick:** Then you could use this enum from above here: 

```tsx
  onSelect={(option) => handleSelect(option, FilterNames.Technology)}
```

Sumick (approved):
Approving, but please do simplify this filter logic, as in my example below. Let's also add enum for filter names, to keep them standarized

PR #28, opened 2023-07-07
fix: refactor views display and add padding

wawrzonnn:


app/(profile)/create-profile/page.tsx
```diff
@@ -16,7 +15,7 @@ const CreateProfilePage = async () => {
   }
 
   return (
-    <div className={styles.main}>
+    <div className={styles.container}>
```
Sumick, line 18: **nitpick:** This should be a separate component, eg: `Container`

src/components/Filters/Filters.tsx
```diff
@@ -88,15 +88,9 @@ const Filters = () => {
...
-        <button className={styles.devTypeBtn + ' ' + styles.frontend}>
-          Frontend
-        </button>
-        <button className={styles.devTypeBtn + ' ' + styles.backend}>
-          Backend
-        </button>
-        <button className={styles.devTypeBtn + ' ' + styles.fullstack}>
-          Fullstack
-        </button>
+        <button className={styles.frontend}>Frontend</button>
```
Sumick, line 91: **nitpick:** I know it's not yours but I also should be separate component :D

app/globals.scss
```diff
@@ -23,11 +23,7 @@ body {
 }
 
 .container {
-  max-width: 1180px;
-  min-height: 100vh;
+  // max-width: 1128px;
```
Sumick, line 26: **nitpick:** Container should always have maxWidth (which is specified in Figma) + it should be centered with `margin: 0 auto`

Sumick (changes requested):
Few fixes

app/(profile)/my-profile/container/Container.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import { PropsWithChildren } from 'react'
+import styles from './Container.module.scss'
+
```
Sumick, line 4: **nitpick:** Why the same component is duplicated? We should have one container for centering web app content

Sumick (approved):
Approving, but please fix this duplicated container before merging

AniaPiWo (approved):

mimakuch (approved):
Looks cool!

PR #30, opened 2023-07-07
feat: add endpoint for getting authorized user data

wawrzonnn:


NerdbordMentor (changes requested):
## Code Review Report

### Task goal
The task was to implement an endpoint for getting authorized user data.

### Requirements
The endpoint should return authorized user data with the following payload schema:
```javascript
interface UserPayload {
   id: string;
   profileId: string;
   githubCredentials: {
     username: string;
     email: string;
   }
}
```
The required endpoint schema is `GET /auth/me`.

### Code changes
Two files were added: `app/api/auth/me.ts` and `app/api/auth/types.ts`.

#### app/api/auth/me.ts
This file implements the required endpoint. It uses the `getSession` function from `next-auth/react` to get the current session. If there is no session or user, it returns a 401 status with an 'Unauthorized' error. If there is a session and user, it constructs a `UserPayload` object from the session data and returns it with a 200 status.

#### app/api/auth/types.ts
This file defines the `UserPayload` interface as required by the task.

### Review
The code changes cover the task requirements. The implemented endpoint returns the authorized user data in the required format. The payload schema is correctly defined and used. The endpoint correctly handles the case where there is no session or user. The code is clean and well-structured. 

However, there is a minor discrepancy between the task requirements and the implemented code. The `UserPayload` interface as defined in the task requirements includes a `githubCredentials` object with `username` and `email` properties, but the implemented code includes only a `username` property in this object. The `email` property is added directly to the `UserPayload` object instead of being included in the `githubCredentials` object. This should be corrected to fully meet the task requirements.

PR #33, opened 2023-07-10
feat: added picture to schema, rebuild create user

DanielSzczucki:


app/api/auth/[...nextauth]/types.ts
```diff
@@ -1,8 +1,8 @@
...
-    id: string
-    profileId: string
-    githubCredentials: {
-      username: string
-      email: string
-    }
-  }
\ No newline at end of file
+  id: string
+  profileId: string
```
Comment (Sumick) [inline app/api/auth/[...nextauth]/types.ts:3]: **nitpick:** `profileId` could be null, because user can exists without it, so: ` profileId: string | null`

app/api/profiles/route.ts
```diff
@@ -21,18 +21,19 @@ export async function GET() {
 
 export async function POST(request: NextRequest) {
   try {
-    const userData: CreateProfilePayload = await request.json()
```
Sumick, line 24: **nitpick:** This is wrong domain - we should not create user in `profile`, but in `user`. This endpoint should be used only for creating profile

prisma/schema.prisma
```diff
@@ -38,7 +38,7 @@ model Profile {
 }
 
 model GitHubDetails {
-  id       String @id @default(uuid())
```
Sumick, line 41: **nitpick:** Why id is removed from here?

prisma/schema.prisma
```diff
@@ -38,7 +38,7 @@ model Profile {
...
+  picture  String
   username String
```
Sumick, line 42: **thought:** We could add `@unique` to the `username`, because it will always be unique (it is on Github)

Sumick (changes requested):
We need few changes here

src/components/Headers/DefaultHeader/DefaultHeader.tsx
```diff
@@ -8,13 +8,20 @@ import AddIcon from '@/assets/icons/AddIcon'
...
   const { data: session, status } = useSession()
-  console.log(session)
   const router = useRouter()
 
   if (status === 'authenticated') {
+    const user = {
+      email: session.user.email || '',
+      image: session.user.image || '',
+      name: session.user.name || '',
+    }
```
Sumick, line 22: **issue:** Creating user should not work like that, it should be created on the backend side in authorization logic, not this way in UI

src/lib/createInitialPartOfUser.ts
```diff
@@ -0,0 +1,20 @@
+export interface userDatafromGH {
+  email: string
+  image: string
+  name: string
+}
+
+export const createInitialPartOfUser = async (
```
Sumick, line 7: **issue:** Creating user cannot take place in UI

Sumick (changes requested):
What you did here is a very dirty hack - user creation must take place is backend, authorization logic, not in UI

PR #59, opened 2023-09-11
Feat/Move-profile-fetching-to-the-server

AniaPiWo:


wawrzonnn (approved):

hansac91 (approved):

src/components/Headers/MyProfileHeader/MyProfileHeader.tsx
```diff
@@ -1,13 +1,22 @@
...
 import React from 'react'
-import { useSession } from 'next-auth/react'
 import Link from 'next/link'
 import styles from './MyProfileHeader.module.scss'
 import logo from '@/assets/images/logo.png'
 import Image from 'next/image'
 
-const MyProfileHeader = () => {
-  const { data: session } = useSession()
+type SessionType = {
```
Sumick, line 7: **nitpick:** We have declared this type already - it's in next-auth.d.ts

src/components/Headers/DefaultHeader/DefaultHeader.tsx
```diff
@@ -1,25 +1,30 @@
...
+    image: string
+    profileId: string | null
+  }
+}
+
+interface DefaultHeaderProps {
+  session: SessionType | null
+}
+
+const DefaultHeader = ({ session }: DefaultHeaderProps) => {
```
Sumick, line 22: **nitpick:** Why passing down this session as prop and not `getServerSession` here in this component directly?

src/components/Headers/DefaultHeader/DefaultHeader.tsx
```diff
@@ -1,25 +1,30 @@
...
+import CreateProfileBtn from '@/components/CreateProfileBtn/CreateProfileBtn'
+import MyProfileBtn from '@/components/MyProfileBtn/MyProfileBtn'
 
-const DefaultHeader = () => {
-  const { status, data: session } = useSession()
+type SessionType = {
+  user: {
+    name: string
+    image: string
+    profileId: string | null
```
Sumick, line 14: **nitpick:** This `profileId` has to be removed from session

app/page.tsx
```diff
@@ -3,16 +3,18 @@ import Filters from '@/components/Filters/Filters'
...
+import { authOptions } from '@/lib/auth'
+
+export default async function Home() {
+  const session = await getServerSession(authOptions)
 
-const Home: React.FC = () => {
   return (
     <FiltersProvider>
-      <DefaultHeader />
+      <DefaultHeader session={session} />
```
Sumick, line 14: **nitpick:** It won't work when getServerSession is called directly?

Sumick (changes requested):
We need to remove profileId from session, as we talked about it during call. So we have to get this other way there in the component.

src/lib/auth.ts
```diff
@@ -38,6 +38,7 @@ export const authOptions: NextAuthOptions = {
 
       token.id = foundUser.id
       token.profileId = foundUser.profile?.id
+      token.profile = !!foundUser.profile
```
Sumick, line 41: **issue:** But we had to remove `profileId` from session because it leads to critical bug. Adding `profile` to session is basically the same problem

src/lib/auth.ts
```diff
@@ -61,9 +62,7 @@ export const authOptions: NextAuthOptions = {
       if (session?.user) {
         session.user.id = token.id as string
         session.user.email = token.email as string
-        session.user.profileId = token.profileId
-          ? (token.profileId as string)
-          : null
+        session.user.profile = token.profile as boolean
```
Sumick, line 65: **issue:** You did basically the same what you removed :P

types/next-auth.d.ts
```diff
@@ -6,11 +6,11 @@ declare module 'next-auth' {
    */
   interface Session {
     user: {
-      profileId: string | null
       id: string
       email: string
       image: string
       name: string
+      profile: Profile | null
```
Sumick, line 13: **issue:** But why? You exchanged profileId for profile which is the same

Sumick (changes requested):
We were supposed to remove profileId from session, not exchange it for whole `profile` object. We can't have profile data in session as there is possibility that it will be null even when user will have profile. We have to remove all profile details from token and session

Sumick (approved):
Looks good! Please fix type name before merging

PR #71, opened 2023-10-06
feat: added user profile github link

wawrzonnn:


hansac91 (approved):

Sumick (changes requested):
User, and Profile relate to each other. You dont need to query them separately. You can include them in each other, and get from DB as a single record.

src/data/frontend/user/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface UserModel {
+  id: string
+  profileId: string
+  githubDetails: {
+    username: string | null
```
Sumick, line 5: **nitpick:** username cannot be null when GighubDetails is created

Sumick (approved):
Looks good!

darionauta (approved):
OK

PR #72, opened 2023-10-06
feat: user roles setup

darionauta:


wawrzonnn (approved):

Sumick (changes requested):
We should not add `role` to Profile, but to User. User can exist without Profile, and role should be always present for User, no matter if User has profile or not.

hansac91 (approved):

Sumick (changes requested):
These are still old changes. `role` is part of Profile, and should be part of User. Btw, can we set `roles` instead of role, so it's possible to hold array of roles instead of one? It will be more proper solution.

Sumick (approved):
Looks good!

PR #73, opened 2023-10-09
feat: refactor profile entity to handle states

darionauta:


src/backend/profile/profile.service.ts
```diff
@@ -87,7 +87,7 @@ export async function createUserProfile(
       seniority: profileData.seniority,
       techStack: profileData.techStack,
       employmentType: profileData.employmentType,
-      isPublished: profileData.isPublished ?? false,
+      state: PublishingState.PENDING,
```
Sumick, line 90: **nitpick:** Should be DRAFT when created

Sumick (approved):
Looks good! I found one thing to fix before merging

PR #74, opened 2023-10-13
Feat/setup dashboard cms

mmcode9407:


wawrzonnn (approved):
Good job! I think it's time to introduce global styles for our headers. We have several common classes like .wrapper, .headerContent, .logo, .title, and .frameButtons that we use in every header. It might be beneficial to consolidate them.

src/components/Headers/DashboardHeader/DashboardHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Button } from '@/components/Button/Button'
+import { MyProfileButton } from '@/components/MyProfileButton/MyProfileButton'
+import logo from '@/assets/images/logo.png'
+
+import styles from './DashboardHeader.module.scss'
+
+const DashboardHeader = async () => {
+  const session = await getServerSession(authOptions)
+
+  const profile = session
```
darionauta, line 21: I wonder if we could shorten this code a bit?
For example:
`const profile = session && await findUserByEmail(session.user.email)`
`if(!profile || !profile?.roles.includes(Role.MODERATOR)) redirect(AppRoutes.home)`

darionauta (approved):

app/(moderation)/dashboard/layout.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  return (
+    <html lang="en">
+      <body className={ibm.className}>
```
Sumick, line 22: **nitpick:** This should not be a root layout, I will explain it in review summary

Sumick (changes requested):
We need to move RootLayout to the root of `app` dir, based on documentation. HTML, body tag, and fonts should be declared only once there. https://nextjs.org/docs/app/api-reference/file-conventions/layout

Then we should create DashboardLayout component, which holds only DashboardHeader, Container, and Children. `AuthProvider` can be in RootLayout

Sumick (approved):
Looks really nice now! Good job 🙌

PR #77, opened 2023-10-18
Fix/fixing header

mmcode9407:


hansac91 (approved):

wawrzonnn (approved):
GJ

src/components/CreateProfileBtn/CreateProfileBtn.tsx
```diff
@@ -1,17 +1,23 @@
...
 
 const CreateProfileBtn = () => {
-  const router = useRouter()
+  const { data: session } = useSession()
+
+  const onClickHandler = async () => {
+    if (!session) {
+      await signIn('github', {
+        callbackUrl: AppRoutes.githubOAuth,
+      })
```
Sumick, line 15: **issue:** user can be logged in without profile. we should `router.push` to create profile page, when user is logged in but does not have profile

Sumick (changes requested):
Looks good, we need one important fix

Sumick (approved):
👌

PR #78, opened 2023-10-18
fix: updated employment type formatting

wawrzonnn:


mmcode9407 (changes requested):
We need to change the value which compares in `checked` properties in every input because `values.employment` after mapping is an one of the string: "Full-time" or "Part-time" or "Contract", not an enum. 
We will need to change the type of employmentType in ProfileModel too.

PR #81, opened 2023-10-21
chore: attribute existence check changed

darionauta:


src/components/renderers/TechnologiesRenderer.tsx
```diff
@@ -0,0 +1,31 @@
+import { ProfileModel } from '@/data/frontend/profile/types'
+
```
Sumick, line 2: **issue:** We should add `use client` here, as it uses `window` - this renderer will crash when rendered on server

Sumick (approved):
Looks good! I found one issue worth addressing before merging

wawrzonnn (approved):

mmcode9407 (approved):
Looks good! Just add border and counter to tabs before merge

PR #85, opened 2023-10-31
Feat/implement profile state management

darionauta:


app/api/profiles/[id]/reject/route.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function POST(
+  request: NextRequest,
+  { params }: { params: { id: string } },
+) {
+  const profileId = params.id
+
+  const session = await getServerSession(authOptions)
+  const user = session && (await findUserByEmail(session.user.email))
+
+  if (!user || !user?.roles.includes(Role.MODERATOR)) {
```
Comment (Sumick) [inline app/api/profiles/[id]/reject/route.ts:17]: **nitpick:** As this check is gonna be used frequently, we can move it to some `utils/auth` function, like:

```ts
requireUserRoles([Role.Moderator])
```

app/api/profiles/[id]/state/route.ts
```diff
@@ -0,0 +1,33 @@
...
+export async function PATCH(
+  request: NextRequest,
+  { params }: { params: { id: string } },
+) {
+  const id = params.id
+
+  const session = await getServerSession(authOptions)
+  const user = session && (await findUserByEmail(session.user.email))
+
+  if (!user || !user?.roles.includes(Role.MODERATOR)) {
```
Comment (Sumick) [inline app/api/profiles/[id]/state/route.ts:18]: **nitpick:** we could reuse role check here

src/contexts/ToastContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const ToastContext = createContext<ToastType | null>(null)
+
+export function ToastContextProvider({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  const [message, setMessage] = useState('')
+  const [toastStatus, setToastStatus] = useState<ToastStatus>(
```
Sumick, line 24: **issue:** This toaster is missing queueing - toasts should stack. Currently they will overwrite themselves

src/contexts/ModalContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <ModalContext.Provider
+      value={{
+        profileId,
+        setProfileId,
+        showRejectModal,
+        setShowRejectModal,
+      }}
+    >
+      {children}
```
Sumick, line 27: **nitpick:** I would suggest considering using React.portal here to render modal. It will prevent uneccessary rerenders

Sumick (changes requested):
Code looks great! I found an issue in Toast Provider, we are missing toast queueing - currently they will overwrite themselves when called one by one

wawrzonnn (approved):
GJ

mmcode9407 (approved):
GJ!

src/contexts/ModalContext.tsx
```diff
@@ -0,0 +1,38 @@
+'use client'
+
+import { PropsWithChildren, createContext, useContext, useState } from 'react'
+
+export type ModalContextType = {
+  profileId: string
+  setProfileId: (id: string) => void
+  showRejectModal: boolean
```
Sumick, line 8: **nitpick:** I think we should only have (of course we we assume that only one modal can be visible at the time, but let's leave it for now:

```ts
export type ModalContextType = {
  isModalVisible: boolean;
  showModal: (state: boolean) => void
}
```

app/(moderation)/layout.tsx
```diff
@@ -9,6 +12,7 @@ export default function DashboardLayout({
 }) {
   return (
     <>
+      <RejectingReason />
```
Sumick, line 15: **nitpick:** Currently we don't know how and where this modal is being called. We should place it where we call it.

src/components/ProfileList/ModerationProfileListItem.tsx
```diff
@@ -13,19 +11,53 @@ import { JobSpecialization } from '@/components/ProfileList/profile-data'
...
 
-const stateStatus = (state: string): React.ReactNode => {
+type StateStatusProps = {
+  profile: ProfileModel
+}
+
+function StateStatus({ profile }: StateStatusProps) {
+  const { id, state } = profile
+  const { addToast } = useToast()
+  const { setProfileId, setShowRejectModal } = useModal()
```
Sumick, line 31: **nitpick:** Instead of keeping profile id in modal provider, just render this Reject Modal here, and pass `profileId` as prop.

Sumick (changes requested):
Almost there! We need last improvements in modals logic

Sumick (approved):
Looks great! Awesome job 🔥👏

PR #86, opened 2023-11-06
feat: implement profile preview

darionauta:


app/(moderation)/dashboard/profile/[id]/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+export default async function ModerationUserProfile({
+  params,
+}: {
+  params: { id: string }
+}) {
+  const profile = await getProfileByUserId(params.id)
+
+  if (!profile || !requireUserRoles([Role.MODERATOR]))
+    redirect(AppRoutes.dashboard)
+
```
Comment (wawrzonnn) [inline app/(moderation)/dashboard/profile/[id]/page.tsx:22]: Should the user be redirected to AppRoutes.dashboard if they're not a moderator? It seems this condition will allow non-moderator access to the dashboard. Is this intended?

src/components/Headers/UserProfileHeader/UserProfileHeader.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default function UserProfileHeader({
+  userProfile,
+  withBackButton,
+}: UserProfileHeaderType) {
+  return (
+    <div
+      className={cx('wrapper', {
+        [styles.withBackBackButton]: !!withBackButton,
+      })}
+    >
```
wawrzonnn, line 19: Take a look at the second image, doesn't it look clearer that way? We define the class names upfront and then apply a single class name to the JSX element 
darek1
darek2

wawrzonnn (approved):
Great job!

src/components/ProfileList/ModerationProfileListItem.tsx
```diff
@@ -101,7 +109,7 @@ export const ModerationProfileListItem: React.FC<{ profile: ProfileModel }> = ({
       <div
         className={styles.container}
         onClick={() =>
-          router.push(`${AppRoutes.userProfile}/${profile.userId}`)
+          router.push(`${AppRoutes.dashboardProfile}/${profile.userId}`)
```
wawrzonnn, line 112: I think we should add the onClick handler to the parent with the 'moderationFrame' class. This way, the entire tile in the UI is clickable, not just half of it.

src/components/AssignRole/AssignRole.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useModal } from '@/contexts/ModalContext'
+import { Button } from '@/components/Button/Button'
+
+import combineClasses from '@/utils/combineClasses'
+import styles from './AssignRole.module.scss'
+import modalStyles from '@/components/Modal/Modal.module.scss'
+import { Inter, IBM_Plex_Sans } from 'next/font/google'
+
+const ibm = IBM_Plex_Sans({ subsets: ['latin'], weight: ['400'] })
+const inter = Inter({ subsets: ['latin'], weight: ['500'] })
```
Sumick, line 11: **nitpick:** Why do we have to duplicate font family declaration each time? I think it should be handled from one place

src/components/Headers/ModerationActionHeader/ModerationActionHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [isAssignModal, showAssignModal] = useState(false)
+
+  return (
+    <div className={styles.wrapper}>
+      <GoBackButton>Go back</GoBackButton>
+      <div className={styles.actionsWrapper}>
+        <Button
+          variant="action"
+          onClick={() => {
+            showAssignModal(true)
```
Sumick, line 25: **nitpick:** Why 2 modals are called here - it does not like best solution, as we duplicate logic that already exists.

Sumick (approved):
Looks good - I've left 2 nits to consider. I wonder if we can optimize modal calling to not duplicate it, if we have already modal provider that should be a main manager for handling modals in our app

wawrzonnn (approved):

src/components/RejectingReason/RejectingReason.tsx
```diff
@@ -39,28 +34,26 @@ export default function RejectingReason({ profileId }: { profileId: string }) {
         ToastStatus.INVALID,
       )
     })
-    showModal(false)
+    closeModal()
   }
 
   if (!isModalVisible) return null
```
Sumick, line 40: **nitpick:** I think we don't need this `if` now

app/(moderation)/dashboard/profile/[id]/page.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  params: { id: string }
+}) {
+  const profile = await getProfileByUserId(params.id)
+
+  if (!profile || !requireUserRoles([Role.MODERATOR])) redirect(AppRoutes.home)
+
+  return (
+    <div className={styles.wrapper}>
+      <Modal />
```
Comment (Sumick) [inline app/(moderation)/dashboard/profile/[id]/page.tsx:25]: **nitpick:** I think we should have this modal only in one place like `ModalProvider` instead of duplicating them around in the system

src/components/AssignRole/AssignRole.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useModal } from '@/contexts/ModalContext'
+import { Button } from '@/components/Button/Button'
+
+import styles from './AssignRole.module.scss'
+import modalStyles from '@/components/Modal/Modal.module.scss'
+
+export default function AssignRole({ profileId }: { profileId: string }) {
+  const { isModalVisible, closeModal } = useModal()
+
+  if (!isModalVisible) return null
```
Sumick, line 11: **nitpick:** I think we don't need this `if` now. We can decouple these modal, and call them eg. `AssignRoleModal`, and pass them eg. `onClose`. It will be easier to manage, and test them

src/contexts/ModalContext.tsx
```diff
@@ -4,19 +4,33 @@ import { PropsWithChildren, createContext, useContext, useState } from 'react'
...
+  showModal: (element: React.ReactNode) => void
+  modalContent: React.ReactNode | null
+  closeModal: () => void
 }
 
 const ModalContext = createContext<ModalContextType | undefined>(undefined)
 
 export function ModalProvider({ children }: PropsWithChildren) {
-  const [isModalVisible, showModal] = useState(false)
+  const [isModalVisible, setIsModalVisible] = useState(false)
```
Sumick, line 15: **nitpick:** I think we don't need it now, as we can check it by checking if `modalContent` is not null

src/contexts/ModalContext.tsx
```diff
@@ -4,19 +4,33 @@ import { PropsWithChildren, createContext, useContext, useState } from 'react'
...
+  const [modalContent, setModalContent] = useState<React.ReactNode | null>(null)
+
+  const showModal = (element: React.ReactNode) => {
+    setModalContent(element)
+    setIsModalVisible(true)
+  }
+
+  const closeModal = () => {
+    setIsModalVisible(false)
```
Sumick, line 24: **nitpick:** let's do `setModalContent(null)`

src/contexts/ModalContext.tsx
```diff
@@ -4,19 +4,33 @@ import { PropsWithChildren, createContext, useContext, useState } from 'react'
...
+  }
 
   return (
     <ModalContext.Provider
       value={{
         isModalVisible,
```
Sumick, line 30: **nitpick**: We can simplify it, or remove this boolean completely

```tsx
   value={{
        isModalVisible: Boolean(modalContent),
```

src/contexts/ModalContext.tsx
```diff
@@ -4,19 +4,33 @@ import { PropsWithChildren, createContext, useContext, useState } from 'react'
...
 const ModalContext = createContext<ModalContextType | undefined>(undefined)
 
 export function ModalProvider({ children }: PropsWithChildren) {
-  const [isModalVisible, showModal] = useState(false)
+  const [isModalVisible, setIsModalVisible] = useState(false)
+  const [modalContent, setModalContent] = useState<React.ReactNode | null>(null)
+
+  const showModal = (element: React.ReactNode) => {
+    setModalContent(element)
+    setIsModalVisible(true)
```
Sumick, line 20: **nitpick:** we don't need handling boolean now

Sumick (approved):
Mostly looks good! We are missing last few improvements, let's fix them before merging:)

hansac91 (approved):

app/layout.tsx
```diff
@@ -19,10 +33,13 @@ export default function RootLayout({
 }) {
   return (
     <html lang="en">
-      <body className={ibm.className}>
+      <body className={commonClasses}>
         <AuthProvider>
           <ToastContextProvider>
-            <ModalProvider>{children}</ModalProvider>
+            <ModalProvider>
+              <Modal />
```
Sumick, line 40: **nitpick:** Why not in ModalProvider?

Sumick (approved):
Looks good! Amazing job 🙌 Leaving one comment

PR #94, opened 2023-11-29
test: add contact developer path test and selector data-test-id in co…

agnbart:
…mponents


cypress/e2e/contact_dev_path.spec.cy.ts
```diff
@@ -0,0 +1,20 @@
+describe('The Home Page', () => {
+  it('Good Dev Local', () => {
+    cy.viewport(1920, 1080)
+    cy.visit('http://localhost:3000/', { failOnStatusCode: false })
```
mmcode9407, line 4: We could use '/' here because we have setted a baseURL in cypress.config.js.

cypress/e2e/contact_dev_path.spec.cy.ts
```diff
@@ -0,0 +1,20 @@
+describe('The Home Page', () => {
+  it('Good Dev Local', () => {
+    cy.viewport(1920, 1080)
```
mmcode9407, line 3: I think we could set it in cypress.config.js as a global option for every test :)

mmcode9407 (approved):
Looks good! I left two comments to consider before merging :)

wawrzonnn (approved):

cypress/e2e/contact_dev_path.spec.cy.ts
```diff
@@ -0,0 +1,20 @@
...
+    cy.wait('@profileLoaded')
+    cy.get('[data-test-id="contactBtn"]').click()
+    cy.get('input[name="senderFullName"]').type('Jan Testowy')
+    cy.get('input[name="senderEmail"]').type('test@mail.server.pl')
+    cy.get('input[name="subject"]').type('Remote job offer')
+    cy.get('textarea[name="message"]').type(
+      'We have a very interesting job offer for you.',
+    )
+    cy.get('[data-test-id="submitBtn"]').click()
+    cy.get('[data-test-id="closeBtn"]').click()
```
Sumick, line 16: **nitpick:** I think we are missing some confirmation here

Sumick (approved):
Mostly looks good, but I think we are missing checking if there is a confirmation afer submitting?

PR #96, opened 2023-11-30
Feat/profile form UI fixes

AniaPiWo:


hansac91 (approved):

wawrzonnn (approved):
FB_IMG_1700672571039

app/(profiles)/my-profile/create/loading.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import styles from './page.module.scss'
+
+const Loading = () => {
```
Sumick, line 4: **nitpick:** I think we should reuse this loader, not duplicate it

Sumick (changes requested):
We should fix this Loader to not duplicate it every time - it should be some one component, and then reused

Sumick (approved):
🔥🔥🔥

PR #99, opened 2023-12-09
Feat/email succesful registration

agnbart:


src/lib/mailersendClient.ts
```diff
@@ -0,0 +1,66 @@
...
+          },
+        ],
+      },
+    ]
+
+    const emailParams = new EmailParams()
+      .setFrom(sentFrom)
+      .setTo(recipients)
+      .setReplyTo(sentFrom)
+      .setSubject('Welcome to Good Dev Hunting')
```
Sumick, line 51: **issue:** Subject should also be dynamic, as we want to send many different emails

src/lib/mailersendClient.ts
```diff
@@ -0,0 +1,66 @@
...
+    const mailerSend = new MailerSend({
+      apiKey: process.env.MAILERSEND_API_KEY!,
+    })
+
+    const sentFrom = new Sender(
+      process.env.MAILERSEND_FROM!,
+      process.env.MAILERSEND_FROM_NAME!,
+    )
+
+    const recipients = [new Recipient(email, name)]
```
Sumick, line 15: **issue:** We should be able to send single email to many recipients

Sumick (changes requested):
Mostly looks good - we need to make 3 values to be dynamic when sending email, and we are ready to go:)

hansac91 (approved):

wawrzonnn (approved):

src/lib/auth.ts
```diff
@@ -56,7 +59,13 @@ export const authOptions: NextAuthOptions = {
           image: castedProfile.avatar_url,
         })
 
-        // TODO(Agnieszka): Send email to user about successful registration
+        if (createdUser) {
+          await mailersendClient.sendMail(
+            [new Recipient(createdUser.email, castedProfile.login)],
+            process.env.MAILERSEND_TEMPLATE_ID!,
```
Sumick, line 65: **nitpick:** We should not keep TEMPLATE_ID, and SEND_SUBJECT in `.env` because we will have a lot of them, and currently it suggests that we have only one. We should create something like, which will be held somewhere in `mailerSendClient.ts`:

```ts
enum TransactionMailTemplateIds {
 welcomeMail = "12355asdgg",
 // then, we will add more template ids later
 ...
}
```

And then:

```ts
 await mailersendClient.sendMail(
            [new Recipient(createdUser.email, castedProfile.login)],
            TransactionMailTemplateIds.welcomeMail,
            "Welcome to Good Dev!",
```

Sumick (changes requested):
We are almost here - one little thing left:)

Sumick (approved):
Good job! 🙌

mmcode9407 (approved):

PR #101, opened 2023-12-12
feat: add GitHub Actions for Cypress tests and code quality checks

wawrzonnn:


hansac91 (approved):

src/components/TextInputWithDropdown/TextInputWithDropdown.tsx
```diff
@@ -7,7 +7,15 @@ import styles from './TextInputWithDropdown.module.scss'
 import { countries } from '@/data/frontend/profile/countries/countries'
 import useOutsideClick from '@/hooks/useOutsideClick'
 
-const TextInputWithDropdown = () => {
+interface TextInputWithDropdownProps {
+  dataTestId?: string
+  dataTestId2?: string
```
Sumick, line 12: **nitpick:** we should not have to pass such props. Only `id` would be enough

yarn.lock
Sumick, line 1: **nitpick:** yarn.lock to remove

src/components/TextInputWithDropdown/TextInputWithDropdown.tsx
```diff
@@ -52,12 +60,14 @@ const TextInputWithDropdown = () => {
         excludeDigits
         onClick={handleCountryInputClick}
         inputRef={inputRef}
+        dataTestId={dataTestId}
```
Sumick, line 63: **nitpick:** No need to pass dataTest this deep, we can do it directly in component

src/components/Dropdowns/DropdownCountry/DropdownCountry.tsx
```diff
@@ -11,9 +11,11 @@ import styles from './DropdownCountry.module.scss'
 const DropdownCountry = ({
   value,
   setIsDropdownActive,
+  dataTestId,
 }: {
   value: string
   setIsDropdownActive: React.Dispatch<React.SetStateAction<boolean>>
+  dataTestId?: string | undefined
```
Sumick, line 18: **nitpick:** We should not need such prop drilling here

src/backend/profile/profile.service.ts
```diff
@@ -189,3 +189,16 @@ export async function getProfileByUserEmail(email: string) {
 
   return null
 }
+
+export async function deleteUserProfileByEmail(email: string) {
+  const user = await prisma.user.findUnique({
+    where: { email },
+    include: { profile: true },
+  })
```
Sumick, line 197: **nitpick:** We should check it earlier:

```ts
 const user = await prisma.user.findUnique({
    where: { email },
    include: { profile: true },
  })

if (!user) {
 throw Error("User not found")
}

await prisma.profile.delete({
      where: { id: user.profile.id },
})
```

app/api/test/profile/route.ts
Sumick, line 1: **nitpick:** Why `/test` path - we should add it normally in `/api/profile/route.ts`

Sumick (changes requested):
I know I have to help here, but can't do it right now - let's do little cleanup first, I will return to it probably tomorrow

PR #109, opened 2023-12-17
Fix/investigate session finding user logic

szymonkadas:
# Summary: 
- User with deleted account can't see nor edit other users profile data. Also user can now login to create account (clicking on createProfile also automatically creates account due to github authorization)
- finding user/profiles by user id's/email are now safekeeped from falsy values. where clause with falsy value returned first value from db. 
- profile pages and components are now also looking at session.user.id (probably) because of src/lib/auth.ts authOptions.callbacks jwt function that returned object with id null. Object that is indeed defined, so hence not falsy.

# Notes:
Problem has probably occurred mainly due to "if" conditions not enough strict for actual data behaviour, and tricky prisma findFirst functionality (undefined values would prove to unexpected behaviours).


wawrzonnn (approved):

mmcode9407 (approved):
Looks good! :)

src/backend/user/user.service.ts
```diff
@@ -18,6 +18,10 @@ export async function getUsersPayload() {
 }
 
 export async function getUserById(id: string) {
+  if (!id) {
```
Sumick, line 21: **nitpick:** We should not need that, as `id` is required

src/backend/user/user.service.ts
```diff
@@ -37,6 +41,10 @@ export async function getUserById(id: string) {
 }
 
 export async function findUserByEmail(email: string) {
+  if (!email) {
+    return null
```
Sumick, line 45: **nitpick:** We should not need that, as `email` is required

src/components/CreateProfileBtn/CreateProfileBtn.tsx
```diff
@@ -16,7 +15,7 @@ const CreateProfileBtn = () => {
   }
 
   const onClickHandler = async () => {
-    if (!session) {
+    if (!session?.user?.id) {
```
Sumick, line 18: **nitpick:** We should leave as it is, and make session null if it's not present

src/components/Headers/AppHeader/AppHeader.tsx
```diff
@@ -19,7 +18,7 @@ const AppHeader = async () => {
     ? await getProfileByUserEmail(session.user.email)
     : null
 
-  if (session) {
+  if (session?.user?.id) {
```
Sumick, line 21: **nitpick:** We should leave as it is, and make session null if it's not present

Sumick (changes requested):
Instead of changing so many things here, we should be able to make it from one place. Maybe just let's return null instead of an object with id of null? Then you would not have to change other things and add more checks
Zrzut ekranu 2023-12-19 o 15 38 54

szymonkadas:
Closing due to too many conflicts from changes from the meantime.

PR #112, opened 2023-12-20
fix: jwt returns null if user cannot be found in db.

szymonkadas:
# Summary: 
- changed JWT next auth interface to type union between extended with id JWT and null
- JWT callback function of AuthOptions returns null if user is not found in db.
- Session callback function of AO now takes into account JWT possibility of null. Solution is up to discussion REQUEST FOR COMMENTS.

# Notes: 
myprofile page imo should have this if statement regardless of choice (edit page has such if statement, and as far as i can see both of them have equal session dependencies).

### DIFF
--- a/app/(profiles)/my-profile/create/page.tsx
+++ b/app/(profiles)/my-profile/create/page.tsx
@@ -1,21 +1,20 @@
-import React from 'react'
-import { getServerSession } from 'next-auth'
-import { authOptions } from '@/lib/auth'
-import { redirect } from 'next/navigation'
-import styles from './page.module.scss'
-import PersonalInfo from '@/components/CreateProfile/PersonalInfo/PersonalInfo'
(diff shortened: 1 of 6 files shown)


app/(profiles)/my-profile/create/page.tsx
```diff
@@ -1,21 +1,20 @@
...
+import { AppRoutes } from '@/utils/routes'
+import { getServerSession } from 'next-auth'
+import { redirect } from 'next/navigation'
+import styles from './page.module.scss'
 
 const CreateProfilePage = async () => {
   const session = await getServerSession(authOptions)
 
-  if (!session) {
+  if (!session || !session.user) {
```
Sumick, line 17: **nitpick:** We can simplify this type of `if`:

```ts
if (!session?.user)
```

Sumick (approved):
Looks nice! Good job on investigating this that deeply 🙌

wawrzonnn (approved):

PR #116, opened 2023-12-21
fix: assign role button and modal are now resembling user roles

szymonkadas:
# Summary: 
- assign role button and modal are now properly displaying information depending on selected user roles.

# Notes: 
- backend logic as i can see is not implemented yet, task idea: implement this feature (as far as i know it should be possible to do so right now)

### DIFF
--- a/app/(moderation)/dashboard/profile/[id]/page.tsx
+++ b/app/(moderation)/dashboard/profile/[id]/page.tsx
@@ -1,13 +1,14 @@
+import { getProfileByUserId } from '@/backend/profile/profile.service'
+import ModerationActionHeader from '@/components/Headers/ModerationActionHeader/ModerationActionHeader'
 import UserProfileHeader from '@/components/Headers/UserProfileHeader/UserProfileHeader'
 import UserProfileDetails from '@/components/UserProfile/UserProfileDetails/UserProfileDetails'
 import UserProfileMain from '@/components/UserProfile/UserProfileMain/UserProfileMain'
-import ModerationActionHeader from '@/components/Headers/ModerationActionHeader/ModerationActionHeader'
-import { AppRoutes } from '@/utils/routes'
-import { redirect } from 'next/navigation'
-import { Role } from '@prisma/client'
 import { requireUserRoles } from '@/utils/auths'
-import { getProfileByUserId } from '@/backend/profile/profile.service'
+import { AppRoutes } from '@/utils/routes'
+import { Role } from '@prisma/client'
+import { redirect } from 'next/navigation'
 
+import { findUserByEmail } from '@/backend/user/user.service'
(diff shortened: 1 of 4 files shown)


app/(moderation)/dashboard/profile/[id]/page.tsx
```diff
@@ -16,12 +17,14 @@ export default async function ModerationUserProfile({
...
+  const user = profile && (await findUserByEmail(profile.userEmail))
 
-  if (!profile || !requireUserRoles([Role.MODERATOR])) redirect(AppRoutes.home)
+  if (!profile || !user || !requireUserRoles([Role.MODERATOR]))
+    redirect(AppRoutes.home)
 
   return (
     <div className={styles.wrapper}>
-      <ModerationActionHeader userProfile={profile} />
+      <ModerationActionHeader userProfile={profile} userRoles={user.roles} />
```
Comment (Sumick) [inline app/(moderation)/dashboard/profile/[id]/page.tsx:27]: **nitpick:** I think you don't have to pass user roles because you can pull it out straight in component using `useSession` hook

src/components/AssignRoleModal/AssignRoleModal.tsx
```diff
@@ -1,16 +1,19 @@
...
   profileId,
+  userRoles,
   onClose,
 }: {
   profileId: string
+  userRoles: Role[]
   onClose: () => void
 }) {
-  return (
+  return !userRoles.includes(Role.MODERATOR) ? (
```
Sumick, line 16: **nitpick:** I think you don't have to pass user roles because you can pull it out straight in component using `useSession` hook

Sumick (approved):
Looks good! One thing to check, because you will probably not have to pass this prop, as you can pull roles from user from `useSession` hook

app/(moderation)/dashboard/profile/[id]/page.tsx
Comment (szymonkadas) [reply app/(moderation)/dashboard/profile/[id]/page.tsx:27]: The user we're working on dashboard/profile:id is not the one in the session. 
In this case it would be the one that owns the profile, hence useSession wouldn't work (cuz it works with moderator that moderates user profiles). 
I could get roles using existent prop (userProfile) but it would mean redundant call in order to obtain these roles once again. 
If such approach would prove to be more beneficial, then we'll keep next commit that I'll provide in few minutes from this comment.

Comment (szymonkadas) [reply app/(moderation)/dashboard/profile/[id]/page.tsx:27]: Okay, nevermind as fas as I can see i'd have to create container server component with call to prisma db to obtain these roles once again only to pass data as children/props down to the client component so it wouldn't make sense I suppose?

Sumick (approved):

wawrzonnn (approved):
Good Job

AniaPiWo (approved):

PR #122, opened 2024-01-08
fix: country & cities are unique. Update works properly. Prisma changes

szymonkadas:
# Description: 
- Country and city tables are now containing only country/city related data, individual attributes (openForRelocations) are moved to profile.
- Country and city updates related to profiles now use proper method (connectOrReplace) so they won't get updated as a whole, instead their relationship with profile is updated respectively.
- Applied necessary type changes and code on frontend

# Notes: 
Added distuingished city & country relocation in this frontend file instead of unified based on country. 
(15th changed file) src/components/UserProfile/UserProfileMain/UserProfileMain.tsx


Sumick (approved):
Good job!

hansac91 (approved):

wawrzonnn (approved):

PR #124, opened 2024-01-09
feat/discord webhooks

szymonkadas:
# Summary:
- implemented webhook instead of bot due to simplier nature of such approach, and lack of need for more advanced usages of dc bot.
- webhook is used in src/lib/discord
- notifications are implemented in user.service and profile.service respectively

# Caveats: 
- notification messages need to be polished (request for comments)
-  DON"T HAVE AN IDEA HOW TO GET CURRENT WEBSITE URL ON LINK MESSAGES, env var?

# Notes: 
When finished we need to create webhook on dc and assign it's URL to new env var e.g. WEBHOOK (now using TEST_WEBHOOK)


src/backend/user/user.service.ts
```diff
@@ -102,6 +103,9 @@ export async function createUser(userDataFromGh: CreateUserPayload) {
       githubDetails: true,
     },
   })
+  await httpClient.post('http://localhost:3000/api/discord', {
```
Sumick, line 106: **issue:** localhost

src/lib/apiClient.ts
```diff
@@ -6,13 +6,19 @@ import {
...
 export const apiClient = {
   publishMyProfile: async (profileId: string) => {
-    const publishedProfile = await httpClient.post<undefined, ProfileModel>(
+    // the function didn't receive ProfileModel, it had received Profile instead (from updateUserData);
+    const publishedProfile = await httpClient.post<undefined, Profile>(
       `/api/profiles/${profileId}/publish`,
     )
+    await httpClient.post('/api/discord', {
+      message: `User ${publishedProfile.fullName} has published profile`,
+      link: `http://localhost:3000/dashboard/profile/${publishedProfile.userId}`,
```
Sumick, line 20: **issue:** such notifications should happen in backend

src/lib/apiClient.ts
```diff
@@ -50,7 +56,11 @@ export const apiClient = {
         PublishingStateData,
         ProfileModel
       >(`/api/profiles/${profileId}/state`, payload)
-
+      // send discord notification
+      await httpClient.post('/api/discord', {
```
Sumick, line 60: **nitpick:** No need to do it client side

app/api/discord/route.ts
```diff
@@ -0,0 +1,26 @@
...
+  try {
+    const data = await request.json()
+    if (!data.message) {
+      console.error('message not provided for discord notification')
+      return new NextResponse('message not provided for discord notification', {
+        status: 500,
+      })
+    }
+    // sending the message through webhook url
+    await httpClient.post(`${process.env.TEST_WEBHOOK}`, {
```
Sumick, line 14: **nitpick:** No need to create API route for that. Such notifications should be called server side, so you can create something like `/lib/discord.ts` where you will have POST to webhook url. Then just use this method when specific events happen, but let's do it server side only

src/backend/user/user.service.ts
szymonkadas: kind off not resolved! (now on profile.service)

szymonkadas:
Updated pr description, last issue to resolve is localhost on linked message, any help how to resolve this?

src/backend/profile/profile.service.ts
```diff
@@ -154,7 +155,13 @@ export async function updateUserData(
     },
     data: userDataToUpdate,
   })
-
+  if (userDataToUpdate?.state) {
+    // I have no idea how to get current app url for link
+    discordNotification(
+      `User's ${updatedUser.fullName} profile has got new status: ${userDataToUpdate.state}! Profile`,
+      `http://localhost:3000/dashboard/profile/${updatedUser.userId}`,
```
Sumick, line 162: **suggestion:** You can do this by putting origin env, so `${process.env.NEXT_PUBLIC_APP_ORIGIN_URL}/dashboard/profile/${updatedUser.userId}`

src/backend/profile/profile.service.ts
```diff
@@ -154,7 +155,13 @@ export async function updateUserData(
     },
     data: userDataToUpdate,
   })
-
+  if (userDataToUpdate?.state) {
+    // I have no idea how to get current app url for link
+    discordNotification(
```
Sumick, line 160: **nitpick:** Let's change this func name to more specific, and action type, eg: `sendDiscordNotificationToWebhook`.

src/lib/discord.ts
```diff
@@ -0,0 +1,15 @@
+import { httpClient } from '@/lib/httpClient'
+
+export async function discordNotification(message: string, link?: string) {
```
Sumick, line 3: **nitpick:** Let's parametrize this function a bit, and make these arguments more descriptive. We also should build message format before we pass it to not hardcode message format. I would also parametrize wehbookURL so we can use this function to send notifications to multiple channels. I would go with something like:

```ts
export async function sendDiscordNotificationToWebhook(params: {
message: string;
webhookUrl: string;
}) {
 await httpClient.post(webhookUrl, {
      content: message,
})
}
```

and if you don't want to copy-paste webhook url each time, then you can do:

```ts
export async function sendDiscordNotificationToModeratorChannel(message: string) {
  sendDiscordNotificationToWebhook({
    webhookUrl: process.env.TEST_WEBHOOK,
    message,
  })
}
```

Sumick (changes requested):
We are almost there! Last polishing and we are good to go

src/backend/profile/profile.service.ts
szymonkadas: Env var it is then, great, thanks for help!

Sumick (approved):
Good job!

hansac91 (approved):

mmcode9407 (approved):

PR #126, opened 2024-01-10
Fix/problem with country flags windows

mr-fox93:


wawrzonnn (approved):

szymonkadas (approved):
Seems good, dunno about yarn mentioned above.

PR #129, opened 2024-01-11
Feat/landing navbar herosec

hansac91:


wawrzonnn (approved):

app/(profiles)/page.tsx
Sumick, line 1: **nitpick:** We should extract this page from our `(profiles)` directory, as landing is not part of profiles anymore

Sumick (changes requested):
Mostly looks good! Few changes to implement:

- [ ] Fonts dont load properly. You can preview your deployment here: https://good-dev-hunting-app-git-feat-landing-navbar-herosec-nerdbord.vercel.app/
Zrzut ekranu 2024-01-11 o 23 28 12

- [ ] Find talents button does not work. it should redirect to `/profiles` where all profiles live
- [ ] let's move out landing `page.tsx` from `(profiles)` dir, and put it in root `./app`

szymonkadas (approved):

app/(profiles)/profiles/page.tsx
Sumick, line 1: **nitpick:** No need to duplicate this `profiles` directory. Let's just change `(profiles)` to `profiles`, and move this `page.tsx` together with `page.module.scss`

app/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+import styles from './page.module.scss'
+import classNames from 'classnames/bind'
+import CreateProfileBtn from '@/components/CreateProfileBtn/CreateProfileBtn'
+import ProfilePicture from '@/assets/images/ProfilePicture.png'
+import Image from 'next/image'
+import FindTalentsBtn from '@/components/FindTalentsBtn/FindTalentsBtn'
+import AppHeader from '@/components/Headers/AppHeader/AppHeader'
+
+const Page: React.FC = () => {
+  const cx = classNames.bind(styles)
```
Sumick, line 12: **nit:** Let's move out `cx` out of component body (outside const)

src/components/FindTalentsBtn/FindTalentsBtn.tsx
```diff
@@ -0,0 +1,20 @@
+'use client'
+import React from 'react'
+import { Button } from '@/components/Button/Button'
+import { AppRoutes } from '@/utils/routes'
+import { useRouter } from 'next/navigation'
+
+const FindTalentsBtn = () => {
+  const router = useRouter()
+
```
Sumick, line 9: **nitpick:** Let's return null if current route is `AppRoute.profiles`, so it does not show when we are currently at this route

Sumick (changes requested):
Few changes to implement:

- [ ] Landing container broke 
Zrzut ekranu 2024-01-14 o 14 06 03

- [ ] Fonts still don't load properly
- [x] we should not show Find talents button in header when we are currently in `/profiles` route
- [ ] we should not duplicate `profiles` directory but change existing directory name from `(profiles)` to `profiles`

hansac91:
Here's my reply:

>  Fonts dont load properly. You can preview your deployment here: https://good-dev-hunting-app-git-feat-landing-navbar-herosec-nerdbord.vercel.app/

I checked the link above and on localhost and everything is okay. 

> * [ ]  Find talents button does not work. it should redirect to `/profiles` where all profiles live - fixed

Sumick (approved):
👌 Let's remove `yarn.lock` before merging, so you don't have this conflict

PR #136, opened 2024-01-18
Feat/landing find next talent section

meczajkowski:


mr-fox93 (commented):
consider redirecting to profiles after clicking on the card and generating random profiles from the database each time. @Sumick we need your opinion :)

wawrzonnn (approved):

mmcode9407 (commented):
Looks good! But consider one thing :) We have a `ModerationProfileListItem` which looks like `ProfileListItem` but has extra buttons.  May can we use a `ProfileCard` there in some way?

Sumick (approved):
Nice! Showing random 6 profiles from DB is a good idea, we can implement it in separate task

meczajkowski:
### Preview





### Notes

> consider redirecting to profiles after clicking on the card and generating random profiles from the database each time. @Sumick we need your opinion :)

done in #138 

> Looks good! But consider one thing :) We have a `ModerationProfileListItem` which looks like `ProfileListItem` but has extra buttons. May can we use a `ProfileCard` there in some way?

I believe the profile card should be refactored in a separate task since it is not responsive due to a fixed width. This component should support widths less than 260px and a maximum of 399px because it is currently causing inconsistencies in the layout and overflow X. 
We can change every profile card in app to the ProfileCard components when it is done.
CC: @Sumick

PR #138, opened 2024-01-19
Fix/landing next talent section

mr-fox93:
! before merge this pr, merge : Feat/landing find next talent section #136  in first place because I have improved the operation of that pr and the TalentSectio.tsx file.

### DIFF
--- a/src/components/landing-page/TalentSection/TalentSection.tsx
+++ b/src/components/landing-page/TalentSection/TalentSection.tsx
@@ -1,162 +1,14 @@
 import React from 'react'
 import styles from './TalentSection.module.scss'
-
-import { ProfileModel } from '@/data/frontend/profile/types'
-import { EmploymentType, PublishingState } from '@prisma/client'
-import { JobSpecialization } from '@/components/ProfileList/profile-data'
-
-// components
 import FindTalentsBtn from '@/components/FindTalentsBtn/FindTalentsBtn'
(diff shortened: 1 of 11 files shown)


meczajkowski (approved):

wawrzonnn (approved):

szymonkadas (approved):

src/components/landing-page/TalentSection/TalentSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+import ProfileCard from '@/components/ProfileCard/ProfileCard'
+import { getPublishedProfilesPayload } from '@/backend/profile/profile.service'
+import Link from 'next/link'
+import { AppRoutes } from '@/utils/routes'
+import { ProfileModel } from '@/data/frontend/profile/types'
+
+const TalentSection = async () => {
+  const profiles = await getPublishedProfilesPayload()
+
+  let randomProfiles: ProfileModel[] = []
```
Sumick, line 13: **issue:** We should not do it this way - we should get random 6 records from DB instead of getting all of them, loading to memory, and filter on FE. Let's create new method eg: `getRandomProfiles(amount: number)`, and take only 4 records from DB.

Sumick (changes requested):
We should not query random records this way - im leaving explanation in comment

PR #141, opened 2024-01-22
Fix/next talent new

mr-fox93:
- add genRandomProfiles(profilesCount: number) in profile.service.ts
- implemented genRandomProfiles() in profile.service.ts, which accepts a
 parameter specifying the number of random profiles to fetch from the database like:
genRandomProfiles(6) show 6 profiles.

### DIFF
--- a/src/components/landing-page/TalentSection/TalentSection.tsx
+++ b/src/components/landing-page/TalentSection/TalentSection.tsx
@@ -1,162 +1,14 @@
 import React from 'react'
 import styles from './TalentSection.module.scss'
-
-import { ProfileModel } from '@/data/frontend/profile/types'
-import { EmploymentType, PublishingState } from '@prisma/client'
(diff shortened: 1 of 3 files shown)


src/components/landing-page/TalentSection/TalentSection.tsx
```diff
@@ -166,8 +18,13 @@ const TalentSection = () => {
         </small>
       </div>
       <div className={styles.talents}>
-        {mockedCards.map((cardData) => (
-          <ProfileCard key={cardData.id} data={cardData} />
+        {profiles?.map((cardData) => (
```
Sumick, line 21: **nit:** No need to optional chaining as it will be always array returned

Sumick (approved):
Good job!

meczajkowski (approved):
Looks good! Organizing imports would be a plus.

AniaPiWo (approved):

PR #146, opened 2024-01-22
fix: multiselect in position work now

mr-fox93:
- Added `handleButtonClick` function in Filters.tsx to enable multi-select functionality for position selection.
- Updated `jobSpecializationFilter` from `string` to `string[]` in FilterContext to support multiple selections.
- Modified `filterByPosition` in filters.ts to accommodate the new multi-select feature.


szymonkadas (approved):
Seems fine, dunno about profile-data.ts, ig it should be deleted entirely?

mr-fox93:
> Seems fine, dunno about profile-data.ts, ig it should be deleted entirely?

line 1-28 is important, I commented the code, which is probably no longer used, because we now show profiles from the database everywhere, not stock ones. @Sumick  take a look please ?

wawrzonnn (approved):
Could you please remove the profile-data file as it seems unnecessary? Just ensure to move the 'enum JobSpecialization' we're using to our frontend types.

mmcode9407 (approved):

src/components/ProfileList/profile-data.ts
```diff
@@ -28,38 +28,38 @@ export enum JobSpecialization {
...
+//     remote: 'Remote',
+//     technology: ['Javascript', 'React', 'Vue.js'],
+//   },
+//   {
+//     id: 2,
+//     name: 'Kristin Watson',
+//     seniority: 'Junior',
+//     jobSpecialization: JobSpecialization.Frontend,
+//     others: '+5 more',
+//     country: 'Poland',
```
Sumick, line 49: **nit:** Why comment? If commenting not removing, let's add explanation in code why it's not removed

Sumick (changes requested):
Let's add this landing fix for negative values when selecting from DB to this PR before merging, because app cannot be deployed now

mr-fox93:
> Let's add this landing fix for negative values when selecting from DB to this PR before merging, because app cannot be deployed now

i think now should work :

export async function getRandonProfiles(profilesCount: number) {
  const totalProfiles = await prisma.profile.count()

  if (profilesCount >= totalProfiles) {
    return getPublishedProfilesPayload()
  }

  const maxSkipValue = totalProfiles - profilesCount
  const skipValue =
    maxSkipValue > 0 ? Math.floor(Math.random() * maxSkipValue) : 0

  const randomRecords = await prisma.profile.findMany({
    take: profilesCount,
    where: {
      state: PublishingState.APPROVED,
    },
    orderBy: {
      id: 'asc',
    },
    skip: skipValue,
    include: {
      user: {
        include: {
          githubDetails: true,
        },
      },
      country: true,
      city: true,
    },
  })

  return randomRecords.map(serializeProfileToProfileModel)
}

src/components/ProfileList/profile-data.ts
mr-fox93: deleted, it was a fragment used when we did not download profiles from the database, I guess.

src/backend/profile/profile.service.ts
```diff
@@ -206,14 +206,25 @@ export async function getProfileByUserEmail(email: string) {
 }
 
 export async function getRandonProfiles(profilesCount: number) {
```
Sumick, line 208: **nit:** `getRandomProfiles*` you have typo `n`

Sumick (approved):
Looks good!

PR #149, opened 2024-01-23
fix: [edit] pic 4mb limit, error msg added when bigger

AniaPiWo:


src/components/UserPhotoUploader/UserPhotoUploader.tsx
```diff
@@ -14,20 +14,27 @@ export const UserPhotoUploader = () => {
   const { imageUploadError, setImageUploadError, setSelectedFile } =
     useUploadContext()
   const fileInputRef = React.useRef<HTMLInputElement>(null)
+  const [errorMsg, setErrorMsg] = useState('')
```
Sumick, line 17: **issue:** We should use Formik errors, not custom one

NerdbordMentor (dismissed):
## Code Review Report

### Review Summary:
The code changes in `UserPhotoUploader.tsx` seem to address the task requirement of providing specific error information when the picture uploaded by the user exceeds the maximum allowed size of 4 MB.

### Code Changes Review:
1. A new state variable `errorMsg` has been introduced to store the error message.
2. The `handleFileChange` function has been updated to include a condition that checks if the file size is less than or equal to 4 MB.
3. If the file size exceeds 4 MB, `setImageUploadError` is set to `true`, and `errorMsg` is set to 'Choose picture smaller than 4MB'.
4. The error message displayed to the user is now dynamic and will show the content of `errorMsg`.
5. An additional error message has been added for cases where the file is not an image or another error occurs during the upload process.

### Feedback on Task Requirements:
- The requirement states that the error message should read "Choose picture max. 4 MB". However, the implemented error message is "Choose picture smaller than 4MB". This is a minor discrepancy in wording but should be addressed to match the exact requirement.
- The code correctly sets an error message when the file size is too large and when there's a different issue with the file upload.

### Suggested Improvements:
To strictly adhere to the task requirements, the error message should be updated to match the specified text exactly. Here's the suggested change:

```diff
- setErrorMsg('Choose picture smaller than 4MB')
+ setErrorMsg('Choose picture max. 4 MB')
```

Additionally, for consistency and clarity, consider updating the other error message to be more specific about the issue, such as indicating that the file is not a valid image type:

```diff
- setErrorMsg('Picture failed to upload, try again')
+ setErrorMsg('Invalid image type, please choose a different picture')
```

### Conclusion:
The code changes cover the task requirements with a minor discrepancy in the error message text. Adjusting the text to match the requirement exactly would complete the task successfully.

Sumick (changes requested):
We use Formik + Yup for validation, we should move it there

mmcode9407 (approved):

Sumick (approved):
🎸

mr-fox93 (approved):

PR #151, opened 2024-01-23
fix: move 'profile not found' to the left side

mr-fox93:
- use devtools to find problem, only delete display:flex


Sumick (approved):
Let's remove other (rest) flex values before merging, as we don't need them anymore

mmcode9407 (approved):
Check gaps on mobile becouse `gap` doesn't work without `flex` 😀

AniaPiWo (approved):

PR #154, opened 2024-01-23
refactor[profiles]/tech as separate model

szymonkadas:
# Summary:
- changed profile.techStack in db so it would be easier to work with it.
- updated all code already related to techStack, especially profile updating chunks.

# Caveats: 
- this is not yet ready to merge code for sure, there's many aspects to discuss, some of them were mentioned in the comments (in code). Request for comments.
- code is yet to be optimised, would love some suggestions  
- not sure if every change was made in good direction, request for comments!


szymonkadas:
Oh and please, i would mostly love suggestions about updating user with tech stack!

prisma/schema.prisma
```diff
@@ -23,27 +23,41 @@ model User {
...
-  techStack                 String[]
+  techStack                 TechnologyOnProfile[]
   employmentType            EmploymentType
-  state                     PublishingState @default(DRAFT)
+  state                     PublishingState   @default(DRAFT)
   rejectionReasons          RejectionReason[]
 }
 
+model Technology {
+  name        String                 @unique
```
szymonkadas, line 48: forgot about @id, name would serve as an id for simplicity

src/components/Filters/Filters.tsx
```diff
@@ -78,17 +52,35 @@ const Filters: React.FC = () => {
...
-  const handleSelectMulti = (option: string): void => {
-    let newFilters: string[]
-    if (technologyFilter.includes(option)) {
-      newFilters = technologyFilter.filter(
-        (selectedOption) => selectedOption !== option,
-      )
-    } else {
-      newFilters = [...technologyFilter, option]
+
+  const handleSelectMulti = (
```
szymonkadas, line 56: changes from other branch, (implementing multi-select on seniority/tech etc)

Sumick (changes requested):
Let's simplify it first, and I will do another review round:

- [ ] Let's simplify DB schema, so we add only one new model which is `Technology`. Let's remove `TechnologyOnProfile` as it adds not necessary layer of complexity. `Technology` model should look like following:
```ts
model Technology {
  id               String.               @default(uuid())
  name        String                 @unique
  profiles     Profile[]
}
```
Then Profile model should be:
```
model Profile {
  // ... all previous fields
  techStack:  Technology[]
}
```

- [ ] Let's keep `ProfileModel` name instead of changing it to `ProfileModelSimplified` as it's not domain related change
- [ ] When user updates techStack in Profile it should create new Technology if it does not exists or if it exists it should connect to existing one
- [ ] We should type reusable includes object using Prisma.validator
- [x] We should not delete Technologies when removing them from Profile, but we should disconnect them from Profile leaving them in database, so other Profile can connect to it

src/backend/profile/profile.service.ts
```diff
@@ -128,33 +105,41 @@ export async function createUserProfile(
...
 ) {
   const updatedUser = await prisma.profile.update({
     where: {
       id,
     },
     data: userDataToUpdate,
   })
+  // update user techStack
+  if (techStack) {
+    await prisma.technologyOnProfile.deleteMany({
```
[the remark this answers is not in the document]
szymonkadas: It's actually removing relationship not the technology itself

szymonkadas:
Didn't type include object, typed objects obtained from prisma calls using this include object.

src/backend/profile/profile.service.ts
```diff
@@ -247,16 +204,34 @@ export async function getRandomProfiles(profilesCount: number) {
...
-      city: true,
-    },
+    include: includeObject,
   })
 
   return randomRecords.map(serializeProfileToProfileModel)
 }
+
+// Due to many to many relationship these 2 functions are kinda necessary to properly update user techstack.
+export async function upsertTechnology(techName: string) {
```
szymonkadas, line 214: To be deleted.

szymonkadas:
Ready for 2nd round, need to merge changes from the meantime.

Sumick (approved):
I will merge now this PR into my & Ania's PR as well as second Szymon PR as we all have changes related to filter logic, and there are tons of conflicts. It will be easier to fix it on one branch

PR #157, opened 2024-01-25
fix: [dev profile] subject in form, some minor css fix

AniaPiWo:


hansac91 (approved):

src/components/ContactForm/ContactForm.tsx
```diff
@@ -73,9 +73,10 @@ export default function ContactForm({
             <TextInput
               label="Subject"
               placeholder="eg. Job offer - let&rsquo;s talk!"
-              value={formik.values.subject}
+              value={formik.values.subject || "Job offer - let's talk!"}
```
Sumick, line 76: **issue:** You don't set initial form value this way - it should be set in initial formik values

Sumick (changes requested):
We should set this subject initial value in initial form values not directly in input

PR #159, opened 2024-01-25
fix: [dev-profile] Mobile / Header objects alignment to the left

Adriann77:


mr-fox93 (approved):
good job, checked on my phone.

hansac91 (approved):

src/components/Headers/UserProfileHeader/UserProfileHeader.module.scss
```diff
@@ -1,7 +1,7 @@
 .wrapper {
   position: absolute;
   display: flex;
-  justify-content: flex-end;
```
mmcode9407, line 4: We style according to the desktop first rule. So move it to the `@include mobile` mixin. The `flex-end` value should stay for divs without `withBackButtonClass` for the desktop.

src/components/Headers/UserProfileHeader/UserProfileHeader.module.scss
```diff
@@ -13,14 +13,20 @@
 }
 
 .withBackBackButton {
-  justify-content: space-between;
```
mmcode9407, line 16: The same as above. Desktop first, and then add `@include Mobile` with some styles for mobile :)

src/components/Headers/UserProfileHeader/UserProfileHeader.module.scss
```diff
@@ -13,14 +13,20 @@
...
-  align-items: center;
-  justify-content: left;
+  align-items: flex-start;
+  justify-content: flex-start;
   flex-wrap: wrap;
   row-gap: 12px;
   column-gap: 10px;
 }
+
+@media (min-width: 768px) {
```
mmcode9407, line 28: When you use the `@include Mobile` mixin with `withBackButton` class, you won't need this rule anymore.

mmcode9407 (changes requested):
Looks good, but we have to fix some styles yet :)

Sumick (approved):

mmcode9407 (approved):

PR #160, opened 2024-01-26
[edit profile] Refactor/multi select availability

szymonkadas:
# Summary:
- implemented multi-select in availability (in profile creation aswell),
- multiselects in edit profile are already integrated here.
- used EmploymentType for this feature.
- EmploymentType in db is now EmploymentType[] (there's a need for db reset on my local db so keep that in mind - i couldn't easily get rid of such need D:)


Sumick (approved):
Im merging to zbiorcza PRka

PR #161, opened 2024-01-27
Feat/[profiles]/db technologies filter

szymonkadas:
# Summary: 
- technology filter is now showing all technologies used by user in db.
- filter shows 8 rows of data (I recommend checking on scss file)

# Notes
- didn't apply any changes to phone display
- as i can see on figma, filters layout is currently kinda off, could try to fix it if needed (would work on phone display aswell)


src/data/frontend/profile/mappers.ts
```diff
@@ -82,15 +84,11 @@ export const mappedLocations = ['Poland', 'Europe', 'Other'].map(
...
-  'Node.js',
-  'React.js',
-  'Vue.js',
-  'Angular',
-  'MongoDB',
-].map((technology) => ({
-  name: technology,
-  value: technology,
-}))
+export const mappedTechnologies: Promise<DropdownOption[]> = apiClient
```
Sumick, line 87: **issue:** mappers are only responsible for mapping data from one type to other type / format - not fetching it

src/app/api/filters/technologies/route.ts
Sumick, line 1: **nit:** we don't need this endpoint as we can use React server component to fetch that data

src/backend/filters/filters.service.ts
Sumick, line 1: **nit:** File name should be the same as domain, so `backend/technology/technology.service.ts`

src/contexts/FilterContext.tsx
```diff
@@ -95,7 +116,20 @@ export const FiltersProvider = ({
   const [availabilityFilter, setAvailabilityFilter] = useState<FilterOption[]>(
     [],
   )
-
+  const [filterLists, setFilterLists] = useState<FilterLists>({
+    technology: [],
+    seniority: mappedSeniorityLevel,
```
Sumick, line 121: **nit:** We can remove this state as we will do it using RSC

src/contexts/FilterContext.tsx
```diff
@@ -95,7 +116,20 @@ export const FiltersProvider = ({
...
+    availability: mappedEmploymentType,
+    location: mappedLocations,
+  })
+  useEffect(() => {
+    mappedTechnologies.then((technologies) =>
+      setFilterLists((prev: FilterLists) => ({
```
Sumick, line 127: **nit:** we can also remove that useEffect

src/lib/apiClient.ts
```diff
@@ -81,4 +81,15 @@ export const apiClient = {
       throw new Error(error as string)
     }
   },
+
+  getTechnologies: async () => {
```
Sumick, line 85: **nit:** we won't need this endpoint, as we will use RSC

Sumick (changes requested):
There is a lot of stuff that we can remove because of using react server component. Just do following, so you can remove all other stuff:


```ts
interface FiltersProps {
   technologies: Technology[]
}

const Filters: React.FC = (props: FiltersProps) => {
  ...
}
```

```ts
const FiltersWithData = async () => {
const technologies = await getTechnologies()
return  <Filters technologies={technologies} />
}
```
```ts
export default function Home() {
  return (
    <FiltersProvider>
      <FiltersWithData />
      <ProfileList />
    </FiltersProvider>
  )
}
```

Sumick (approved):
Very nice:)

wawrzonnn (approved):

hansac91 (approved):

PR #162, opened 2024-01-28
Feat/profiles  db location filter

szymonkadas:
# Summary:
- implemented countries display for location filter based on countries in db.
- unified mapTechnology & mapCountries to mapOption
# Notes:
- Based on this branch mentioned in this pr: https://github.com/nerdbord/good-dev-hunting-app/pull/161


src/data/frontend/profile/mappers.ts
```diff
@@ -52,6 +53,14 @@ export const mapSpecialization = (specialization: JobSpecialization) => {
   }
 }
 
+export const mapOptions = <T extends { name: string }>(
+  items: T[],
```
Sumick, line 57: **praise/thought:** Nice, however we have to be careful, as sometimes we will need for example `item.id` as value instead of name. Or `name` will be formatted, so it's nicely displayed to user but value will be in ugly format that we will use in database (like enums)

Sumick (approved):
Looks good! One thing to consider

mmcode9407 (approved):
👍

mr-fox93 (approved):
if Sumick approved, I have nothing to add 🥇

PR #165, opened 2024-01-29
feat: save contact requests in db

szymonkadas:
# Summary: 
- implemented contactRequest model in db, i've added profileId for relationship. 
- used api endpoint for creating contact request in db, i couldn't figure out a way to do this using server component.

### DIFF
--- a/src/backend/contactRequest/contactRequest.service.ts
+++ b/src/backend/contactRequest/contactRequest.service.ts
@@ -1,0 +1,13 @@
+import { ContactFormRequest } from '@/components/ContactForm/schema'
+import { prisma } from '@/lib/prismaClient'
+export async function saveContactRequest(contactRequest: ContactFormRequest) {
+  return prisma.contactRequest.create({
+    data: {
+      subject: contactRequest.subject,
+      message: contactRequest.message,
+      senderFullName: contactRequest.senderFullName,
+      senderEmail: contactRequest.senderEmail,
+      profileId: contactRequest.profileId,
+    },
+  })
+}
--- a/src/components/ContactForm/ContactForm.tsx
+++ b/src/components/ContactForm/ContactForm.tsx
@@ -4,6 +4,7 @@ import TextArea from '@/components/TextArea/TextArea'
 import TextInput from '@/components/TextInput/TextInput'
 import { ProfileModel } from '@/data/frontend/profile/types'
 import { useAsyncAction } from '@/hooks/useAsyncAction'
+import { apiClient } from '@/lib/apiClient'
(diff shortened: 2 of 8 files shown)


src/backend/contactRequest/contactRequest.service.ts
Sumick, line 1: **nit:** directory names and such service names should be named with kebab-case, so `contact-request`

src/components/ContactForm/ContactForm.tsx
```diff
@@ -23,8 +24,12 @@ export default function ContactForm({
     runAsync(async () => {
       try {
         // Handle submit actions
-        // console.log('Handle submit', values)
-        showSuccessMsg()
+        apiClient.saveContactRequest({
```
Sumick, line 27: **nit**: you should be able to call server action here, let's do this instead of calling endpoint

Sumick (changes requested):
Let's replace API endpoint, and call to server action. We should create new directory for profile related actions, and then make a file there which will hold implementation for that, so: `src/actions/profile/sendContactRequestToProfile.ts`. Remember that you need to add `'use server'` directive at the top of file contents. Check other actions for reference. Then let's just use it in our submit

src/actions/contact-request/saveContactRequest.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const saveContactRequest = async (
+  contactFormRequestData: ContactFormRequest,
+) => {
+  try {
+    const createdContactRequest = await createContactRequest(
+      contactFormRequestData,
+    )
+    if (createdContactRequest) {
+      return true
```
Sumick, line 13: **nit:** Let's just return `createdContactRequest` instead of bool

Sumick (approved):
Nice! One little thing before merging 👌

mmcode9407 (approved):
👍

mr-fox93 (approved):

src/actions/contact-request/saveContactRequest.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const saveContactRequest = async (
+  contactFormRequestData: ContactFormRequest,
+) => {
+  try {
+    const createdContactRequest = await createContactRequest(
+      contactFormRequestData,
+    )
+    if (createdContactRequest) {
+      createdContactRequest
```
Sumick, line 13: **nit:** it's missing return, and let's remove booleans as discussed on other PR

Sumick (commented):
One comment before merging

PR #166, opened 2024-01-29
Feat/spam prevention

szymonkadas:
# Summary:
- implemented spam prevention feature with Ip tracking - one ip shouldn't be able to send 2 messages to one profile, there's also createdAt field so it would be easier to implement deleting such track after some time.
- there's a question about realtionship to profile, should it be implemented? It would be easier to manage IpTrack profileID's list (e.g. profile is deleted, hence ipTrack profileID is deleted aswell)
- didn't test it yet properly (on localhost request headers aren't existent as far as i can see), cuz it's
 --A: working on not integrated yet branch with related feature (save contact request) 
 --B: save contact request +  DB changes, didn't want to push them to remote base.

# Could be useful:
https://stackoverflow.com/questions/68338838/how-to-get-the-ip-address-of-the-client-from-server-side-in-next-js-app (TL;DR request-ip mentioned and NextRequest mentioned)
https://github.com/vercel/next.js/discussions/34609 (TL;DR comment in this thread says: it should work on remote host)

# Notes:
- there are 2 ways of collecting ip, one requires package and is using other interface in request, second one is fully compatible with our current codebase (not sure which one is better, cuz couldn't measure it's performance -> can't get ip so...)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "react-dom": "18.2.0",
         "react-icons": "^4.9.0",
         "react-select": "^5.7.3",
+        "request-ip": "^3.3.0",
         "sass": "^1.63.4",
         "yup": "^1.2.0"
       },
@@ -34,6 +35,7 @@
         "@types/node": "20.2.5",
         "@types/react": "^18.2.48",
         "@types/react-dom": "^18.2.4",
+        "@types/request-ip": "^0.0.41",
         "@typescript-eslint/eslint-plugin": "^5.59.7",
         "classnames": "^2.3.2",
         "cypress": "^13.6.0",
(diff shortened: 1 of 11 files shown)


Sumick (changes requested):
As Im investigating that problem now, I read that keeping IPs in database is not so obvious as it's strictly regulated, and we should inform users that we do this. As this IP prevention would be just temporary solutions before we implement new user type, let's just do validation for now based on sent `ContactRequest`. 

- [x] Let's remove all IP related logic
- [x] Let's add validation, so when user submits `ContactRequest` it checks if `senderEmail` already contacted this developer. If yes, throw error that 'You've already contacted this developer'. If no, create new `ContactRequest`

szymonkadas:
Due to almost whole pr being reverted at this point i've recreated this feature on this pr:
https://github.com/nerdbord/good-dev-hunting-app/pull/167

PR #169, opened 2024-01-30
Feat/mailerlite

szymonkadas:
# Summary:
- implemented mailerlite
- user when trying to contact user is added to subscribe group of contractors
- user on registration is added to subscribe group of developers
- mailerlite logic is setupped on backend/contactRequest cuz of it's nature, but i'm not so sure about this placement.


src/backend/contactRequest/contactRequest.service.ts
```diff
@@ -0,0 +1,30 @@
+import { httpClient } from '@/lib/httpClient'
+import { mailerLiteClient } from '@/lib/mailerliteClient'
+
+export async function addSubscriberToMailerLite(
+  email: string,
+  groupId: string,
+  fields?: Record<string, string>,
+) {
```
Sumick, line 8: **nit:** these external lib related methods should be placed in `lib/mailerliteClient.ts`

src/components/ContactForm/ContactForm.tsx
```diff
@@ -18,13 +20,13 @@ export default function ContactForm({
...
 }) {
   const { runAsync, loading } = useAsyncAction()
-
   const handleSendEmail = (values: ContactFormValues) => {
     runAsync(async () => {
       try {
         // Handle submit actions
         // console.log('Handle submit', values)
-        showSuccessMsg()
+        addSubscriberToMailerLite(values.senderEmail, contactGroup),
```
Sumick, line 28: **nit:** this also should be called in `action` from your PR with creating contact request, let's merge first, then we will adjust it

src/lib/mailerliteClient.ts
```diff
@@ -0,0 +1,9 @@
+import MailerLite from '@mailerlite/mailerlite-nodejs'
+
+export const mailerLiteClient = new MailerLite({
+  api_key: process.env.NEXT_PUBLIC_MAILERLITE_KEY || '',
+})
+
+export const devGroup = '111547703431792014'
```
Sumick, line 7: **nit:** this should be an enum as you won't know how many other consts are available, as it will grow, so:

```ts
enum MailerLiteGroups {
  devs = '111547703431792014',
  lookingForDevs = '111547722557818681',
}
```

Sumick (changes requested):
Few changes mostly related to placing new files

src/actions/contact-request/saveContactRequest.ts
```diff
@@ -45,6 +46,14 @@ export const saveContactRequest = async ({
         await deleteSavedContactRequest(senderEmail, profileId)
         throw Error('Failed to send contact request')
       }
+      try {
+        await mailerliteClient.addSubscriberToMailerLite(
```
Sumick, line 50: **nit:** no need for creating another try catch block. Let's put it into previous `try`

src/lib/mailerliteClient.ts
```diff
@@ -0,0 +1,40 @@
...
+      })
+    } catch (error) {
+      console.error('Error adding subscriber to MailerLite:', error)
+      return error
+    }
+  },
+  fetchSubscribersFromMailerLite: async (groupId: string) => {
+    try {
+      return await httpClient.get(
+        `https://connect.mailerlite.com/api/groups/${groupId}/subscribers`,
```
Sumick, line 33: **nit:** i think we can remove it as we dont use it

Sumick (approved):
Good job, 2 small tweaks before merging

PR #170, opened 2024-01-31
feat: added server action responsible for sending mail on CR

szymonkadas:
# Summary: 
- implemented sending mail on contact request using mailersend (in server component)

# Issues: 
- not tested yet! I didn't have api key yet, although results are promising because of 401 (unauthorised).


src/actions/mailing/contactRequestEmail.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type ContactRequestEmailParams = {
+  senderEmail: string
+  senderFullName: string
+  recipientEmail: string
+  subject: string
+}
+
+export const sendContactRequest = async (
+  cREParams: ContactRequestEmailParams,
```
Sumick, line 13: **nit:** It's really hard to guess what it means, let's just do `contactRequestParams` instead of `cREP`. Btw, you can always do:

```ts
export const sendContactRequest = async (
  {senderEmail, senderFullName, subject}: ContactRequestEmailParams,
```

so you don't have to keep that argument name:D

src/components/ContactForm/ContactForm.tsx
```diff
@@ -24,6 +25,12 @@ export default function ContactForm({
       try {
         // Handle submit actions
         // console.log('Handle submit', values)
+        await sendContactRequest({
```
Sumick, line 28: **nit:** This should be fired from this `action` that is on your previous PR. Firstly, contact request should be created, then if successfully created, emails should be sent

Sumick (approved):
Looks good, few notes here - for sure we need to merge this PR with contact request creation first, then adjust these changes to that PR changes. Can you ensure that value of the form is being sent correctly with that email, and is it possible to respond to it directly? There is a chance that template is not necessary form that kind of email

wawrzonnn (approved):

mr-fox93 (approved):
take a look at  Sumick's comments before the merge.

src/components/ContactForm/ContactForm.tsx
szymonkadas: Done.

src/actions/mailing/contactRequestEmail.tsx
szymonkadas: Thanks! Implemented.

PR #171, opened 2024-01-31
feat: implemented sending profile status email on approval/rejection

szymonkadas:
# Summary:
- implemented sending profile status email on rejection/approval.

# Issues:
Not tested yet, didn't have api key.

# Notes:
- merging to feat/mailersendContactRequest (last pr), cuz changes are in the same field.
- feature is working in api endpoint, due to updating profile status nature.


src/actions/mailing/sendProfileStatusEmail.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { mailersendClient, MailTemplateId } from '@/lib/mailersendClient'
+import { PublishingState } from '@prisma/client'
+import { Recipient } from 'mailersend'
+
+export const sendProfileStatusEmail = async (
+  email: string,
+  status: keyof typeof PublishingState,
+) => {
+  function setTemplateByStatus(status: keyof typeof PublishingState) {
```
Sumick, line 11: **issue:** This should not be a single method that manages different statuses, because logic can differ a lot between them. We should make explicit actions, like: `sendProfileApproval`, and `sendProfileRejection` because `sendProfileRejection` needs to get data about rejection reason that is created in DB, and it should send it to the user so he knows why he was rejected

src/app/api/profiles/[id]/state/route.ts
```diff
@@ -20,6 +19,17 @@ export async function PATCH(
   try {
     const userDataToUpdate: PublishingStateData = await request.json()
     const updatedProfileState = await updateUserData(id, userDataToUpdate)
+    const userData = await getUserById(id)
+    if (userData) {
```
Comment (Sumick) [inline src/app/api/profiles/[id]/state/route.ts:23]: **nit:** No need for this logic - if you implement my first comment, let's just do:

```ts
  const updatedProfile = await updateUserData(id, userDataToUpdate)

if (updatedProfile.state === PublishingState.APPROVED) {
     sendProfileApprovalEmail(updatedProfile.email)
}
```

Rejecting should be done in other place (`api/profiles/[id]/reject/route.ts`). We should send reject reason upon rejecting profile

src/app/api/profiles/[id]/state/route.ts
```diff
@@ -20,6 +19,17 @@ export async function PATCH(
...
+      const updatedProfileStateMail = await sendProfileStatusEmail(
+        userData.email,
+        userDataToUpdate.state,
+      )
+    } else {
+      throw new Error(
+        `Couldn't send notification email, userData is: ${userData}`,
```
Comment (Sumick) [inline src/app/api/profiles/[id]/state/route.ts:30]: `nit`: no need for this err

Sumick (changes requested):
We need to split how emails are being sent + handle rejection properly, as it's not implemented correctly now

src/app/api/profiles/[id]/reject/route.ts
```diff
@@ -1,30 +1,44 @@
-import { saveRejectingReason } from '@/backend/profile/rejection.service'
+import { sendProfileRejectedEmail } from '@/actions/mailing/sendProfileStatusEmail'
```
Comment (Sumick) [inline src/app/api/profiles/[id]/reject/route.ts:1]: **nit:** we should not import from `actions` to `service`. Only from `service` to `actions`. If you are missing service method just create `backend/mailing/mailing.service.ts` where you will keep reusable mailing pieces. Then you can import it here or to `actions`

src/app/api/profiles/[id]/reject/route.ts
```diff
@@ -1,30 +1,44 @@
...
     return new NextResponse(null, { status: 401 })
   }
 
   try {
-    const rejectingReasonText = await request.json()
-    const createdReason = await saveRejectingReason(
-      profileId,
-      rejectingReasonText.reason,
-    )
+    const { email, reason } = await request.json()
```
Comment (Sumick) [inline src/app/api/profiles/[id]/reject/route.ts:22]: **nit:** Why passing email if you can just `findProfileById` and get the most safe way

src/components/RejectingReasonModal/RejectingReasonModal.tsx
```diff
@@ -1,19 +1,21 @@
...
 import { ToastStatus, useToast } from '@/contexts/ToastContext'
+import styles from './RejectingReason.module.scss'
 
 export default function RejectingReasonModal({
   profileId,
+  userEmail,
   onClose,
 }: {
   profileId: string
+  userEmail: string
```
Sumick, line 18: **nit:** no need to pass it from frontend, as email is related to profile, and reject reason is part of the profile

src/lib/apiClient.ts
```diff
@@ -58,7 +58,7 @@ export const apiClient = {
   },
   createRejectionReason: async (
     profileId: string,
-    payload: RejectionReason,
+    payload: RejectionReason & { userEmail: string },
```
Sumick, line 61: **nit:** to remove as well

src/actions/mailing/sendProfileStatusEmail.tsx
```diff
@@ -0,0 +1,35 @@
+'use server'
+
+import { mailersendClient, MailTemplateId } from '@/lib/mailersendClient'
+import { Recipient } from 'mailersend'
+
+export const sendProfileApprovedEmail = async (email: string) => {
+  const templateId = MailTemplateId.profileApprovedNotification
+
+  await mailersendClient.sendMail({
+    recipients: [new Recipient(email)],
```
Sumick, line 10: **nit:** these methods should stay in `backend/mailing/mailing.service.ts` as we won't send them from client. Btw, such reusable logic should always stay in service. Then it should be used in actions if needed

Sumick (changes requested):
Starting to look good, we need few improvements:

- [x] move mailing logic to `backend/mailing/mailing.service.ts`
- [ ] Then reuse it in actions if needed, but we don't need mailing in actions as we always send it from other server actions
- [x] remove additional userEmail passing from client, as we can get it directly on backend using profileId

PR #172, opened 2024-02-01
feat: team members are now listed on meet our team, todo: figure out …

szymonkadas:
# Summary:
team members are visible in meet our team

### Caveats:
- I'm not so sure if the way i rendered out meet our team is good enough,
- should vertical card even allow placeholder data?  Should it only display ONLY Team members, and if they're not existent display NOTHING? 

### TODO: 
figure out how to display availability (dots, etc)


src/actions/team/fetchTeam.tsx
```diff
@@ -0,0 +1,12 @@
+'use server'
+
+import { getTeamProfiles } from '@/backend/profile/profile.service'
+
+export async function fetchTeam() {
```
Sumick, line 5: **nit:** Actions are for POSTS, we don't have to keep them here, as you can just use `getTeamProfiles` directly in server component

src/components/landing-page/MeetTeam/MeetTeam.tsx
```diff
@@ -1,33 +1,56 @@
...
 import VerticalCard from '@/components/VerticalCard/VerticalCard'
+import { ProfileModel } from '@/data/frontend/profile/types'
+import { useEffect, useMemo, useState } from 'react'
+import styles from './MeetTeam.module.scss'
 
 const MeetTeam = () => {
+  const [teamData, setTeamData] = useState<ProfileModel[]>([])
+  const content = useMemo(() => {
+    console.log(teamData.length)
+    if (teamData.length > 0) {
```
Sumick, line 12: **issue:** This is super complicated, and you created `server action` to run it on the client which makes no sense. It should look like following (remember to remove `use client` directive). Don't worry about scenario when there are no records. There will always be some records

```tsx
const MeetTeam = async () => {
  const teamProfiles = await getTeamProfiles();

  // Calculate midpoint differently to ensure one half has at least 4 items
  let midpoint;
  if (teamProfiles.length < 8) {
    // Ensure the first half has at least 4 items, if total less than 8
    midpoint = 4;
  } else {
    // Standard calculation if enough items for both halves to have at least 4
    midpoint = Math.ceil(teamProfiles.length / 2);
  }

  const firstHalfProfiles = teamProfiles.slice(0, midpoint);
  const secondHalfProfiles = teamProfiles.slice(midpoint);

  return (
    <section id="MeetTeam" className={styles.wrapper}>
      <div className={styles.titleBox}>
        <p className={styles.sectionName}>WORKING AS A TEAM</p>
        <h2 className={styles.title}>Meet our team</h2>
        <h5 className={styles.subtitle}>Meet passionates behind the scene</h5>
      </div>
      <div className={styles.cardsWrapper}>
        <div className={styles.slider}>
          <div className={styles.cardsBox}>
            {firstHalfProfiles.map(profile => <VerticalCard {...profile} />)}
          </div>
        </div>
        {!!secondHalfProfiles.length && <div className={styles.slider}>
          <div className={styles.cardsBox}>
            {secondHalfProfiles.map(profile => <VerticalCard {...profile} />)}
          </div>
        </div>}
      </div>
    </section>
  );
};

```

Sumick (changes requested):
We need to improve it as it is implemented incorrectly - I've posted solution in comment

Sumick (approved):
Looks good!

mmcode9407 (approved):

hansac91 (approved):

PR #174, opened 2024-02-02
Feat  landing copies

AniaPiWo:


Sumick (approved):
Nice, I think it's ok for now - if anything comes to our minds we will add more if necessary

wawrzonnn (approved):

PR #178, opened 2024-02-03
feat: move uplaodImage to SA; some fixes

mmcode9407:
- I've moved upload file route handler to server action;
- I've added uploadImage function in CreateProfileForm. It has missed;
- I had to use FormData for pass a file to server action. I've found out that the server actions only accept plain object;
- I've removed `apiClient` becouse we won't use it anymore;


wawrzonnn (approved):

szymonkadas (approved):
Looks good

src/app/(auth)/_actions/uploadImage.ts
```diff
@@ -0,0 +1,25 @@
+﻿'use server'
+
```
Sumick, line 2: **nit:** uploading is not a part of `auth` domain. Let's do `app/(files)/_actions/uploadImage.ts`

Sumick (approved):
Looks good, let's just move out `uploadImage` from `auth` domain before merging, details in comment

CharlesSquirel (approved):

PR #199, opened 2024-02-07
Feat/contact request captcha

szymonkadas:
# Summary:
- implemented captcha on client side and server-side. 
- fixed checkbox input text alignment
obraz
obraz


szymonkadas:
GPT instructions:
To add server-side verification of Captcha to this component, you need to:
1. Send the value received from the handleCaptchaChange method to your server.
2. Inside your server-side code, send a POST request to https://www.google.com/recaptcha/api/siteverify with your secret key and the response (the value you received from your front-end).
Let's update the handleCaptchaChange function to send the value to our server:
jsx
const handleCaptchaChange = async (value: string | null) => {
  if (value) {
    try {
      const response = await fetch('/api/verify-captcha', {
        method: 'POST',
        headers: {
          'Content-Type': 'application/json',
        },
        body: JSON.stringify({ captcha: value }),
      });

      const data = await response.json();

      if (data.success) {
        props.onChange(props.name, true);
      }
    } catch (error) {
      console.error('There was an error verifying the reCAPTCHA', error);
    }
  }
  setVerificationComplete(true);
}
On your server, you need to create an endpoint (I'm using Next.js):
js
// pages/api/verify-captcha.js

export default async (req, res) => {
  if (req.method !== 'POST') {
    return res.status(405).end();
  }

  const { captcha } = req.body;

  try {
    const response = await fetch(`https://www.google.com/recaptcha/api/siteverify?secret=${process.env.RECAPTCHA_SECRET_KEY}&response=${captcha}`, {
      headers: {
        "Content-Type": "application/x-www-form-urlencoded; charset=utf-8"
      },
      method: "POST"
    });
    const captchaValidation = await response.json();

    if (captchaValidation.success) {
      return res.status(200).json({ success: true });
    }

    return res.status(422).json({ success: false });
  } catch (error) {
    return res.status(422).json({ success: false });
  }
};
Remember: Never expose your reCAPTCHA secret key to your client side. It's a secret and should be kept on the server, that's why we are making client to send the value to server and then server validates it by making a call to [google.com](http://google.com/) to verify it, in turn keeping the secret key a secret.
Also, note that I'm using fetch here which is not supported in all environments. You may have to use a library like axios or node-fetch to make requests from your server, depending on your environment.

src/app/(profile)/(components)/ContactForm/ContactForm.tsx
```diff
@@ -147,6 +149,15 @@ export default function ContactForm({
               </span>
             </CheckboxInput>
           </InputFormError>
+          <InputFormError error={formik.errors.captcha}>
+            <CaptchaCheckbox
+              id="captcha"
+              label="I am not a robot"
```
Sumick, line 155: **nit:** Let's change copy to `Confirm your humanity with a quick click`

Sumick (approved):
Good job! Let's make few little improvements before merging

src/app/(profile)/_actions/captchaValidation.ts
```diff
@@ -0,0 +1,21 @@
+'use server'
+import { withSentry } from '@/utils/errHandling'
+
+export const captchaValidation = withSentry(async (captchaValue: string) => {
```
Sumick, line 4: **nit:** Let's make its name more action type, so `validateCaptcha`

src/app/(profile)/(components)/ContactForm/ContactForm.tsx
szymonkadas: quite funny label 👀

Sumick (approved):

src/app/(profile)/(components)/ContactForm/schema.ts
```diff
@@ -33,4 +35,7 @@ export const validationSchema = Yup.object().shape({
   terms: Yup.boolean()
     .required('Agreement is required')
     .oneOf([true], 'Agreement is required'),
+  catcha: Yup.boolean()
```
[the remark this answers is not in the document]
szymonkadas: Thanks!

Issue #201, opened 2024-02-07
Login failure after changing primary GitHub email address


ekrzeptowski:
## Description
After changing primary email address, I get the following error message:
`Invalid prisma.user.create() invocation:`
`Unique constraint failed on the fields: ('username')`

## Steps to reproduce
1. Go to https://github.com/settings/emails and change your primary email address.
2. Logout and try to login again with the new email address.
3. See the error message.

Sumick:
It should work now @ekrzeptowski :-) Thank you!

Sumick:
Fix merged

PR #202, opened 2024-02-08
Fix/changed email issue

szymonkadas:
# Summary: 
- implemented double check for existing userData in our database on signIn and jwt in Auth (if user has changed their username or email (BUT NOT BOTH), his data will be updated based on 2nd remaining parameter, if both were updated user will have to create new account).
- this pr solves issue: https://github.com/nerdbord/good-dev-hunting-app/issues/201

# Notes:
- I'm not sure about helper functions placement in our codebase, and their naming request for comments.

### DIFF
--- a/src/app/(auth)/auth.ts
+++ b/src/app/(auth)/auth.ts
@@ -27,8 +27,6 @@ export const authOptions: NextAuthOptions = {
   ],
   callbacks: {
     async jwt({ token, user }) {
-      console.log('TOKEN: ', token)
-      console.log('USER: ', user)
       const foundUser =
         token && token.email && token.name
           ? await findUserByGithubCredentials(token.name, token.email)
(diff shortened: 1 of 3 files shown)


src/app/(auth)/auth.ts
```diff
@@ -28,7 +28,9 @@ export const authOptions: NextAuthOptions = {
   callbacks: {
     async jwt({ token, user }) {
       const foundUser =
-        token && token.email ? await findUserByEmail(token.email) : null
+        token && token.email && token.name
+          ? await findUserByGithubCredentials(token.name, token.email)
```
Sumick, line 32: **nit:** Then just:

```ts
const existingUser = token && token.email && token.name && await syncGithubCredentialsWithUser({
username: token.name,
email: token.email.
})
```

src/app/(auth)/auth.ts
```diff
@@ -41,7 +43,10 @@ export const authOptions: NextAuthOptions = {
       const castedGithubDetails = githubDetails as GitHubProfileAuthed
 
       if (user && githubDetails) {
-        const foundUser = await findUserByEmail(castedUser.email)
+        const foundUser = await findUserByGithubCredentials(
```
Sumick, line 46: then same here: 

```ts
const existingUser = await syncGithubCredentialsWithUser({
username: token.name,
email: token.email.
})
```

Sumick (changes requested):
We need few improvements here - you will find all details below

Sumick (approved):

PR #204, opened 2024-02-08
fix: filtering by techstack only from approved profiles

wawrzonnn:


szymonkadas (approved):

src/backend/technology/technology.service.ts
```diff
@@ -1,8 +1,15 @@
 import { prisma } from '@/lib/prismaClient'
+import { PublishingState } from '@prisma/client'
 
-// redundant name
 export async function getTechnologies() {
```
Sumick, line 4: **nit:** let's change to `getPublishedTechnologies`

Sumick (approved):
One comment before merging

PR #208, opened 2024-02-09
fix: auth fix

Sumick:
Currently logging in was not working as it was trying to overwrite existing unique values


szymonkadas:
Shouldn't it be linked to the issue? Or the 209 pr (the big one?)

PR #212, opened 2024-02-10
Fix problem with enter in mdx

mr-fox93:
Use ThematicBreak from MDX docs, made own css style so break is invisible now.We can add own style like margin. User add ThematicBreak from toolbar.


Sumick:
Im closing as we implemented different solution for now

PR #216, opened 2024-02-11
fix: resolved nesting and SVG attribute errors to fix hydration issue

wawrzonnn:


mr-fox93 (approved):
It seems that the problem has disappeared, in the console I only see information about obsolescence? objectFit. I read that layout="fill" should be added in this version of Next.

image

Sumick (approved):
🫶

AniaPiWo (approved):

PR #218, opened 2024-02-12
feat: now profiles in route profiles show up randomly

mr-fox93:


wawrzonnn (approved):

src/backend/profile/profile.service.ts
```diff
@@ -10,9 +10,10 @@ export async function getPublishedProfilesPayload() {
     },
     include: includeObject.include,
   })
-  const serializedProfile = publishedProfiles.map(
-    serializeProfileToProfileModel,
-  )
+
+  const shuffledProfiles = publishedProfiles.sort(() => Math.random() - 0.5)
```
Sumick, line 14: **nit:** We should do this shuffle in ProfilesList component instead of service logic

Sumick (changes requested):
Let's separate this shuffle logic from backend service

Sumick (approved):

PR #223, opened 2024-02-14
Fix/button styles

hansac91:
1. Improved button styles according to Figma
2. Fixed "Go back" button in mobile version
3. Changed layout when there's one socialItem
4. Fixed SocialItem components in my-profile preview


wawrzonnn (changes requested):
Great job! Please do two more things within the scope of this task:

The goBack button should receive props and direct us either to the homepage or to moderation, depending on where it is currently rendered.
The connect with nerdbord button should be hidden as it is currently causing errors. We will fix this soon.

wawrzonnn (approved):

Sumick (approved):

PR #231, opened 2024-02-20
feat: server action for views counter

SzDabrowski:


src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const countProfileViews = withSentry(
+  async (profileId: string, payload: Prisma.ProfileUpdateInput) => {
+    const foundProfile = await getProfileById(profileId)
+
+    if (!foundProfile) {
+      throw new Error('Profile not found')
+    }
+
+    payload.viewCount = foundProfile.viewCount + 1
```
Sumick, line 16: **nit:** we should not mutate object this way but we should use prisma update query, where we have increment syntax: https://www.prisma.io/docs/orm/reference/prisma-client-reference#atomic-number-operations

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,24 @@
...
+
```
Sumick, line 17: we should validate that when authorized user is owner of the profile it should not be incremented

PR #233, opened 2024-02-21
Feat profile views counter

SzDabrowski:


src/app/(profile)/profiles/[username]/page.tsx
```diff
@@ -44,11 +47,22 @@ const UserProfilePage = async ({
...
   if (!selectedProfile) {
     redirect(AppRoutes.profiles)
   }
 
+  if (!session?.user) {
+    redirect(AppRoutes.profiles)
+  }
+
+  try {
+    await countProfileView(selectedProfile.id, session.user.id)
```
Comment (Sumick) [inline src/app/(profile)/profiles/[username]/page.tsx:61]: **nit:** Let's add return here, because as I understand it returns updated profile?

src/app/(profile)/my-profile/(components)/ProfileViews/ProfileViews.tsx
```diff
@@ -0,0 +1,15 @@
+import ViewIcon from '@/assets/icons/ViewIcon'
+
+import styles from './ProfileViews.module.scss'
+
+const ProfileViews = ({ ViewCount }: { ViewCount: number }) => {
```
Sumick, line 5: **nit:** props names should be named with `camelCase`

Sumick (approved):
Good job! Im leaving two small comments to implement before merging

src/app/(profile)/profiles/[username]/page.tsx
Comment (SzDabrowski) [reply src/app/(profile)/profiles/[username]/page.tsx:61]: Yes, it can return updated profile data. I skipped it because its not passed anywhere. Data presented to user in 'myprofile' is loaded from db.  I will add this if necessary

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,15 @@
...
+  getProfileById,
+  incrementProfileViewCountById,
+} from '@/backend/profile/profile.service'
+import { withSentry } from '@/utils/errHandling'
+
+export const countProfileView = withSentry(
+  async (profileId: string, userId: string) => {
+    const foundProfile = await getProfileById(profileId)
+
+    if (profileId !== userId) {
```
Sumick, line 11: **issue:** This condition is wrong, `profileId` is totally different thing than `userId`

src/app/(profile)/profiles/[username]/page.tsx
```diff
@@ -44,11 +47,22 @@ const UserProfilePage = async ({
   params: { username: string }
 }) => {
   const selectedProfile = await getProfileByGithubUsername(params.username)
+  const session = await getServerSession(authOptions)
 
   if (!selectedProfile) {
     redirect(AppRoutes.profiles)
   }
 
+  if (!session?.user) {
```
Comment (Sumick) [inline src/app/(profile)/profiles/[username]/page.tsx:56]: **issue:** we should not add this redirect, as currently previewing profiles is open for public

Sumick (changes requested):
This will not work - condition is wrong, and there is redirect added that should not be added. We should check session in server action, and then check if found profile owner is matching user id from session.

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -1,14 +1,17 @@
+'use server'
 import {
   getProfileById,
   incrementProfileViewCountById,
 } from '@/backend/profile/profile.service'
 import { withSentry } from '@/utils/errHandling'
 
 export const countProfileView = withSentry(
-  async (profileId: string, userId: string) => {
+  async (profileId: string, userProfileId?: string) => {
```
SzDabrowski, line 9: changed props names, and set userProfileId to be optional

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -1,14 +1,17 @@
...
     const foundProfile = await getProfileById(profileId)
 
-    if (profileId !== userId) {
+    if (profileId === userProfileId) {
+      throw new Error("User can't incerement own profile view count")
+    } else {
```
SzDabrowski, line 14: non logged user can now inceremt view count

src/app/(profile)/profiles/[username]/page.tsx
```diff
@@ -53,12 +53,8 @@ const UserProfilePage = async ({
     redirect(AppRoutes.profiles)
   }
 
-  if (!session?.user) {
-    redirect(AppRoutes.profiles)
-  }
-
```
Comment (SzDabrowski) [inline src/app/(profile)/profiles/[username]/page.tsx]: non logged user can view profiles

wawrzonnn (approved):

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,18 @@
...
+  incrementProfileViewCountById,
+} from '@/backend/profile/profile.service'
+import { withSentry } from '@/utils/errHandling'
+
+export const countProfileView = withSentry(
+  async (profileId: string, userProfileId?: string) => {
+    const foundProfile = await getProfileById(profileId)
+
+    if (profileId === userProfileId) {
+      throw new Error("User can't incerement own profile view count")
```
Sumick, line 13: **issue:** We should not throw error here, because it's not an error but intended behaviour to not count own page view

Sumick: let's just call `const updatedProfile = await incrementProfileViewCountById(profileId)` when `profileId !== viewerProfileId`

src/app/(profile)/_actions/countProfileView.ts
Sumick: but you pass `session?.user.id` here which is not `userProfileId`. You need to change you `if` because it will not work.

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,18 @@
+'use server'
+import {
+  getProfileById,
+  incrementProfileViewCountById,
+} from '@/backend/profile/profile.service'
+import { withSentry } from '@/utils/errHandling'
+
+export const countProfileView = withSentry(
+  async (profileId: string, userProfileId?: string) => {
```
Sumick, line 9: it is not `userProfileId` - let's change it to `userId` because you pass `userId` not `profileId`

src/app/(profile)/_actions/countProfileView.ts
```diff
@@ -0,0 +1,18 @@
...
+    const foundProfile = await getProfileById(profileId)
+
+    if (profileId === userProfileId) {
```
Sumick, line 12: it should be `if (foundProfile.id !== userId)` then increment page view

Sumick (changes requested):
It will still not work as your condition is incorrect. You compare `profileId` with `userId` which are totally different things. We also should not throw error when user views own profile

Sumick (approved):
Looks good! 🔥

PR #234, opened 2024-02-21
[Login Page]: Create login page UI

meczajkowski:
This page is currently for UI purposes only and is not linked elsewhere. You can access it by navigating to the `/login` route.

- Implemented the UI for the login page.
- Added a new UI component called Box, which serves as a wrapper component with a fixed width and a lighter background.
- Resolved the issue with the checkbox width.
- Updated the Github Login Button component to support an optional custom label.

_Mobile view adjustments will be implemented in another pull request._

image


src/app/(auth)/login/(components)/LoginHeader/LoginHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+import MyProfileBtn from '@/app/(profile)/(components)/MyProfileBtn/MyProfileBtn'
+import CreateProfileBtn from '@/app/(profile)/my-profile/(components)/CreateProfileBtn/CreateProfileBtn'
+import GithubStarsButton from '@/components/Button/GitHubStarsBtn'
+import { Container } from '@/components/Container/Container'
+import Logo from '@/components/Logo/Logo'
+
+const LoginHeader = async () => {
+  const session = await getServerSession(authOptions)
+
+  const user = session?.user?.email
```
Sumick, line 16: **nit:** I think we can make it reausable function that gets our DB user from session or returns null if not found. We have something similar in `app/(auth)/auth.ts` called `authorizeUser`. I think we can refactor this function, as Im not sure we use it anywhere

src/app/(auth)/login/(components)/LoginHunter/LoginHunter.tsx
```diff
@@ -0,0 +1,47 @@
...
+          tooltipText=" Lorem ipsum dolor sit amet, consectetur adipisicing elit."
+          addImportantIcon={true}
+          placeholder="eg. peter.parker@oscorp.com"
+          onChange={() => {
+            console.log('')
+          }}
+        />
+        <div style={{ marginBlock: '24px' }}>
+          <CheckboxInput
+            id={'terms'}
```
Sumick, line 32: **nit:** Let's move duplicated keys strings into some enum or const like we have in every form currently

Sumick (approved):
Good job! Im leaving 2 small improvements to consider before merging

src/app/(auth)/login/(components)/LoginHeader/LoginHeader.tsx
meczajkowski: implemented as `getAuthorizedUser()`

PR #250, opened 2024-02-28
Feat/filters url and searchbar

wawrzonnn:


src/components/Dropdowns/DropdownFilterMulti/DropdownFilterMulti.tsx
```diff
@@ -78,29 +86,47 @@ export const DropdownFilterMulti = ({
...
+              />
+            )}
+            {filteredOptions.length < 1 && (
+              <p className={styles.noMatchingOptions}>
+                No matching options found
+              </p>
+            )}
+            {options
+              .filter((option) =>
+                option.name.toLowerCase().includes(searchTerm.toLowerCase()),
```
Sumick, line 108: **nit:** We need to split it into smaller components

src/app/(profile)/my-profile/(components)/EditProfileForm/mappers.ts
```diff
@@ -30,4 +30,5 @@ export const mapProfileModelToEditProfileFormValues = (
   githubUsername: profile.githubUsername,
   state: profile.state,
   viewCount: profile.viewCount,
+  userEmail: profile.userEmail,
```
Sumick, line 33: **issue:** We are missing migration file with this new column

Sumick (changes requested):
We are missing migration file here + let's split this search input as separate component. it git big now

PR #256, opened 2024-03-06
Feat/filters in url

mmcode9407:


Sumick (changes requested):
Mostly looks really good - we need few UX improvements and to implement few code improvements

- [x] Filters now close when choosing one 
- [x] Availability filter does not persist selected value when choosing one
- [x] Search input lose focus when typing value
- [x] filter dropdown should close when clicking on it again (when it's open). currently it closes and opens again which is pretty annoying

src/app/(profile)/(components)/Filters/TabFilters.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </SpecializationTab>
+      {specializations.map((spec) => {
+        const color = jobSpecializationThemes[spec.value as JobSpecialization]
+        return (
+          <SpecializationTab
+            key={spec.value}
+            count={numberOfProfiles?.[spec.value] || 0}
+            color={color}
+            href={`${AppRoutes.profiles}/${spec.value}?${new URLSearchParams(
+              searchParams.toString(),
```
Sumick, line 67: **nit:** this URL building logic repeats - let's move it to some function or const, as it's quite complex

src/app/(profile)/profiles/[position]/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+  filterByLocation,
+  filterBySeniority,
+  filterByTechnology,
+} from '@/app/(profile)/(components)/ProfileList/filters'
+import { type JobOfferFiltersEnum } from '@/app/(profile)/types'
+import { getPublishedProfilesPayloadByPosition } from '@/backend/profile/profile.service'
+import Loader from '@/components/Loader/Loader'
+import { Suspense } from 'react'
+
+export default async function PositionPage({
```
Comment (Sumick) [inline src/app/(profile)/profiles/[position]/page.tsx:14]: **nit:** We have to be consistent in terms of domain - let's make it `SpecializationPage`, and `/profiles/[specialization]/page.tsx`. I know we have `position` in DB but we will have to change it for sure

src/app/(profile)/profiles/[position]/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const filteredProfiles = profiles
+    .filter(filterBySeniority(searchParams.seniority?.split(',')))
+    .filter(filterByLocation(searchParams.location?.split(',')))
+    .filter(filterByTechnology(searchParams.technology?.split(',')))
+    .filter(filterByAvailability(searchParams.availability?.split(',')))
+    .filter(filterByFullName(searchParams.search))
+
+  return (
+    <Suspense key={JSON.stringify(searchParams)} fallback={<Loader />}>
```
Comment (Sumick) [inline src/app/(profile)/profiles/[position]/page.tsx:31]: Why such key?

src/utils/createFiltersObjFromSearchParams.ts
```diff
@@ -0,0 +1,20 @@
+﻿import { JobOfferFiltersEnum } from '@/app/(profile)/types'
+import { type ReadonlyURLSearchParams } from 'next/navigation'
+
+export const createFiltersObjFromSearchParams = (
+  searchParams: ReadonlyURLSearchParams,
```
Sumick, line 5: **nit:** This is filter related thing - it should be places somewhere in `app/(profile)/types`. We should try to not import anything from `app/(domain)` to the root directories

Sumick (approved):
Amazing job, great refactor, a lot of improvements - congrats! Im leaving few minor things before merging to staging 🔥🔥🔥

src/app/(profile)/profiles/[position]/page.tsx
Comment (mmcode9407) [reply src/app/(profile)/profiles/[position]/page.tsx:31]: Referring to this part of docs: [https://react.dev/reference/react/Suspense#resetting-suspense-boundaries-on-navigation](url)
Next does not recognize search params changing as a route changing so it does not retrigger the suspense fallback but this solution was for case without having filters inside `layout.tsx`. Now, I can even remove all suspense boundary from the pages 😁

wawrzonnn (approved):

PR #257, opened 2024-03-09
feat: added action for unpublishing profile and implemented it

SzDabrowski:


src/app/(profile)/_actions/unpublishProfile.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!foundProfile) {
+    throw new Error('Profile not found')
+  }
+
+  const updatedProfile = await updateProfileById(foundProfile.id, {
+    state: PublishingState.DRAFT,
+  })
+
+  await sendDiscordNotificationToModeratorChannel(
+    `User's **${updatedProfile.fullName}** profile has got new status: **${updatedProfile.state}**! [Show Profile](${process.env.NEXT_PUBLIC_APP_ORIGIN_URL}/moderation/profile/${updatedProfile.userId})`,
```
Sumick, line 24: **nit:** Let's change copy to `**${updatedProfile.fullName} has unpublished profile**`

src/app/(profile)/(components)/TogglePublishButton/TogglePublishButton.tsx
```diff
@@ -23,10 +24,17 @@ export const TogglePublishButton = (props: TogglePublishButtonProps) => {
   const { loading, runAsync } = useAsyncAction()
 
   const handleButtonClick = async () => {
-    await runAsync(async () => {
-      await publishProfile(profileId)
-      setShowPopup(true)
-    })
+    if (state == PublishingState.APPROVED) {
```
Sumick, line 27: **issue:** This will not work correctly as we have also `PublishingState.REJECTED`. Let's change this function to:

```ts
const handleButtonClick = async () => {

  switch (state) {
      case PublishingState.APPROVED:
       await runAsync(async () => {
        await unpublishProfile(profileId)
        setShowPopup(true)
      })
      return
      case PublishingState.DRAFT:
       await runAsync(async () => {
        await publishProfile(profileId)
        setShowPopup(true)
      })
      return
      case PublishingState.REJECTED:
       await runAsync(async () => {
        await publishProfile(profileId)
        setShowPopup(true)
      })
      return
      default:
      throw Error("Wrong profile action")
   }

}
```

Sumick (changes requested):
Mostly looks good! One issue to resolve + one small improvement

wawrzonnn (approved):

Sumick (approved):
Looks good!

PR #266, opened 2024-03-29
chore: move profiles to client

Sumick:
Moved Profiles fetching to the provider so it's called once in the context of whole app to greatly reduce amount of DB queries


meczajkowski (approved):
Great job!

PR #270, opened 2024-04-09
fix: it was enought to remove line height in active tabs

mr-fox93:
It was enough to remove line height in active tabs

image

image


Sumick (approved):
Code looks good

PR #271, opened 2024-04-09
fix: problem with ridirecting to wrong url when opening profile in mod view

mr-fox93:


Sumick (changes requested):
It won't work properly because we use these cards in other places in the system, and now it's hardcoded and it will always lead to moderation panel

mr-fox93 (commented):
fix this, use usePathname to riderect only in moderation view

Sumick (changes requested):
That's a very bad idea, as you are hardcoding route in dumb component - what if we would like to add different routes? You would have to add a lot of ifs - thats not the way

PR #273, opened 2024-04-13
feat: added clear btn and counter in dropdown

hansac91:


Sumick (approved):
Code looks good! Great job Hania 🔥 I found few minor CSS improvements to implement before merging:

- [ ] there is no gap between Technology(0), and counter (0) [screen]
- [ ] Padding right does not affect right button (Apply) - it should not be sticked do the right as on image
- [ ] Btw - don't we have reset button for desktop? I know we do not have design for that but maybe you can propose some solution for that?:)

Zrzut ekranu 2024-04-15 o 16 05 36

mmcode9407 (approved):
👍

mr-fox93 (approved):
good job !

PR #274, opened 2024-04-14
chore: architecture improvements

Sumick:
Golden rules of this architecture setup:

1. `_actions` should be called from their related models from `_model`
2. `_actions` are always returning model as result of computation
4. Models replace `serializers` functionality, and encapsulate proper `_actions` for their own domain
5. Models implement their own related Prisma interfaces to keep our schema synced with current app data structures
6. Model is now sharable between frontend, and backend as it can be used client side, and server side
7. We should not import `backend/services` logic straight to the components now. We can do it only through `_actions` where serialization to model happens
8. `backend/serializers` are removed now, as models possessed this functionality
9. `_providers` are added now. Providers provide models fetched server side so they can be accessed across client components while optimizing requests 

So, the overview would be (order of importing, it should not be imported reverse way):

`Backend/Service --> (module)/_actions --> (module/_model) --> (module/(components)) [SSR]`
`Backend/Service --> (module)/_actions --> (module/_model) --> (module/_providers) --> (module/(components)) [CSR]`

That makes our backend logic agnostic, and we keep Next.js specific features encapsulated in specific layers while leaving models also framework agnostic


meczajkowski (approved):
Great job! Additionally, I suggest relocating our pages to the (routes) directory. 
For instance, the signin page, currently located at `src/app/(auth)/signin/page.tsx`, could be moved to `src/app/(auth)/(routes)/signin/page.tsx` for better organization and clarity.

wawrzonnn (approved):

mmcode9407 (approved):
Great!

PR #278, opened 2024-04-16
fix: use withStatestatus to redirect

mr-fox93:


Sumick (changes requested):
URL is still hardcoded in our dumb component, we should fix it - currently we cannot predict, and control what is going when clicking on this Card

PR #279, opened 2024-04-16
Fix cards in moderation view does not show editable profiles

mr-fox93:
I removed the Link from ProfileCard, moved the handling to functions in the ProfileList and ModerationProfileListItem files where it redirects us to the appropriate route, i.e. moderation or the standard profile view.


Sumick (changes requested):
We can't remove plausible event tracking + few minor fixes

PR #284, opened 2024-04-22
Fix salary display and save in edit mode

mr-fox93:
image


meczajkowski (approved):

src/app/(profile)/profile.mappers.ts
```diff
@@ -175,7 +175,8 @@ export const mapSpecializations = (
 
 //HOURLY RATE
 export const hourlyRateOptions: DropdownOption[] = [
-  { name: '< 100 zł/h', value: '0-99' },
+  { name: '< 50 zł/h', value: '10-49' },
```
Sumick, line 178: **nit:** Value should be "1-50"

Sumick (approved):
Looks nice! One small thing to implement before merging

src/app/(profile)/profile.mappers.ts
Sumick: Ok, so shoudn't we change other values as well? I mean `50-99`, `100-149` etc

mr-fox93: I will fix it in the branch with the filter because it will be closed after this branch

PR #286, opened 2024-04-22
Fix users with salary should displayed first

mr-fox93:


Sumick (changes requested):
We should not do this this way as it's not a part of filtering logic - I will take this task, as Im not sure now how it should have been done. I have to add new migrations to DB anyway so I will open new PR

PR #281, opened 2024-04-17
Fix redirect card problem v2

mr-fox93:

PR #289, opened 2024-04-23
feat: hide talents name when unlogged

Wiecek-K:


src/app/(profile)/(components)/ProfileCard/ProfileCard.tsx
```diff
@@ -85,9 +87,20 @@ const ProfileCard = ({
...
+              {isHiddenName ? (
+                <div className={styles.nameCover}>
+                  <div className={styles.paddlockContainer}>
+                    <div className={styles.paddlockTop}></div>
+                    <div className={styles.paddlockBottom}></div>
+                  </div>
+                  Login to access
+                </div>
+              ) : null}
+            </div>
```
mmcode9407, line 103: We need to make it as usual. I mean you have to copy the svg from figma and create icon component eg. `PaddlockIcon.tsx` inside `src/assets/icons`. Then, use it as normal component against the paddlock divs. Also, you can return `higlightText` when `isHiddenName` is false againts null.

src/app/(profile)/(components)/ProfileList/ProfileList.tsx
```diff
@@ -2,10 +2,12 @@
 import { ProfileListItem } from '@/app/(profile)/(components)/ProfileList/ProfileListItem'
 import { useProfiles } from '@/app/(profile)/_providers/Profiles.provider'
 import { AppRoutes } from '@/utils/routes'
+import { useSession } from 'next-auth/react'
 import styles from './ProfileList.module.scss'
 
 const ProfileList = () => {
   const { filteredProfiles: profiles } = useProfiles()
+  const { data: session } = useSession()
```
mmcode9407, line 10: I think, you can do this inside `ProfileListItem` to avoid prop drilling. You can do the same with href 👍

mmcode9407 (commented):
I leave some comments 😁

src/app/(profile)/(components)/ProfileList/ProfileList.tsx
```diff
@@ -20,13 +22,25 @@ const ProfileList = () => {
...
+          session?.user ? (
+            <ProfileListItem
+              key={profile.id}
+              data={profile}
+              href={`${AppRoutes.profile}/${profile.githubUsername}`}
+            />
+          ) : (
+            <ProfileListItem
+              key={profile.id}
+              data={{
```
Sumick, line 35: **nit:** Let's pass props explicitly because not it's not clear what is getting into this component

Sumick (approved):
Looks good! One comment from me before merging

src/app/(profile)/(components)/ProfileCard/ProfileCard.tsx
Sumick: Agree!

Sumick:
Zrzut ekranu 2024-04-23 o 17 25 03

Let's also `align-items: center` here, because not it's not aligned correctly

src/assets/icons/PadlockIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+    >
+      <rect
+        x="5"
+        y="11"
+        width="14"
+        height="10"
+        rx="1"
+        stroke="#A687FF"
+        stroke-linecap="round"
+        stroke-linejoin="round"
```
mmcode9407, line 18: Chnage the names on camelCase to avoid errors.

src/assets/icons/PadlockIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+      />
+      <path
+        d="M7 8C7 5.23858 9.23858 3 12 3V3C14.7614 3 17 5.23858 17 8V11H7V8Z"
+        stroke="#A687FF"
+        stroke-linecap="round"
+        stroke-linejoin="round"
```
mmcode9407, line 24: The same

src/app/(profile)/(components)/ProfileCard/ProfileCard.tsx
```diff
@@ -9,20 +9,22 @@ import {
...
-import { type UrlObject } from 'url'
 import styles from './ProfileCard.module.scss'
 
 interface ProfileCardProps {
   onClick?: (e: React.MouseEvent) => void
   data: ProfileModel
   withStateStatus?: boolean
   searchTerm?: string | null
-  href: string | UrlObject
+  isHiddenName?: boolean
```
mmcode9407, line 27: For what is this prop? You already check `session` for displaying the names. If so you should check the `session` inside parent component and then pass the value inside `ProfileCard` and then do not check the `session` inside `PrfoileCard`

mmcode9407 (approved):
I leave some comments before merging :)

src/app/(profile)/(components)/ProfileCard/ProfileCard.tsx
Wiecek-K: bcs we use this component at home page
and Jakub told me that: on the main page we should display full names regardless of whether the user is logged in.
That's why I'm created that prop. Other way to decide when we will hide full names is check the url path but I don't have idea how to do that in server side component.
Now I see, the name is off the mark, I'll extend it. "isHiddenNameWhenUnlogged" will be okay?

mmcode9407: Aaa okay :) but still consider checking the session inside `ProfileListItem` and pass the value of isHiddenName prop based on the session, e.g: `isHiddenName = {session && session.user ? true : false}`. Thanks to that we'll still have a `ProfileCard` as a dump component :)
The name of prop is clear for me. I was thinking why you use it but now it is clear for me :)

src/app/(profile)/(components)/ProfileCard/ProfileCard.tsx
```diff
@@ -70,9 +71,10 @@ const ProfileCard = ({
   const getTechnologyClasses = cx({
     [styles.technology]: true,
   })
+
   return (
     <Link
-      href={href}
+      href={`${AppRoutes.profile}/${data.githubUsername}`}
```
meczajkowski, line 77: This approach would block us getting into moderation view of profile. check #281  and `src\app\(profile)\(components)\ProfileList\ModerationProfilesWithFilter.tsx`

meczajkowski (changes requested):
One thing to fix

meczajkowski (approved):
🥳

PR #295, opened 2024-05-06
Task/refactor headers in app

Wiecek-K:
I merged all header components into one to which we pass in props which variant of buttons to display.


mr-fox93 (approved):
looks good to me, the search bar in the mobile view does not work, I enter the name and it does not find the profile. Other than that it works fine.

mr-fox93:
> > looks good to me, the search bar in the mobile view does not work, I enter the name and it does not find the profile. Other than that it works fine.
> 
> Do you mean the search bar on the /profiles page? If yes - it's weird because it works correctly on my end. image


sorry, my fault, i think this search bar is for 'name' :D but its for search in select.

wawrzonnn (approved):
Great Job Więcek

Sumick (approved):
Good job 🤙

PR #296, opened 2024-05-08
Chore: auto accept profile when work information changed

meczajkowski:
## Summary

- Changed the querying method for profile from querying by email to querying by user ID.
- Modified the `hasProfileValuesChanged` function to no longer query the database, as the complete profile is already available. It now returns an array of changed fields, which correspond to keys in the `ProfileModel`.
- Implemented a new helper function to determine whether the changed fields in a profile are those that would trigger a change to the profile's state to "PENDING."
- Refactored `hasProfileValuesChanged` into a helper function to accept profile changes when work information is modified.
- Adjusted the logic to handle profile state changes based on the modified fields.
- Created a new type `EditProfileFormFields` to represent the fields that can be edited in the profile form.

If we want to exclude additional fields from triggering a change in the profile state, we simply need to remove them from the `profilePendingFields` constant. This provides flexibility in customizing which fields should be considered when determining whether the profile state should change to "PENDING".


mmcode9407 (approved):
🔥

Sumick (approved):
Good job! Looks really good

PR #300, opened 2024-05-09
fix: change logic in technologyfilter

hansac91:
I modified the logic of the technologyFilter to show only records that meet all selected values, rather than just some of them. Essentially, I changed the filter logic to use "AND" for the selected values in the dropdown, instead of "OR"."


Sumick (approved):
🔥

Wiecek-K (approved):

PR #305, opened 2024-05-13
Feat finall setup i18n

mr-fox93:
- use [https://next-intl-docs.vercel.app/](url)
- (profile) and (auth) go to [locale] to handle routing for different languages.
- translation files are located in the messages folder (en.json and pl.json)
- to add a new route in the future, add the appropriate path in the middleware.ts file.
- we add translations in .json files, import them using useTranslations for client comp and getTranslations for async


src/app/[locale]/(auth)/(routes)/signin/(components)/LoginDev/LoginDev.tsx
```diff
@@ -0,0 +1,17 @@
+// components
+import { GithubLoginButton } from '@/app/[locale]/(auth)/(components)/GithubLoginButton/GithubLoginButton'
+import Box from '@/components/Box/Box'
+import { useTranslations } from 'next-intl'
+
+const LoginDev = () => {
+  const t = useTranslations('loginDev')
```
Comment (Sumick) [inline src/app/[locale]/(auth)/(routes)/signin/(components)/LoginDev/LoginDev.tsx:7]: **nit:** We should have these translation namespaces delcared in some enum so it's for example: 

```const t = useTranslations(I18n.SignIn)```

Sumick (changes requested):
There are missing translations on following pages:

- [x] Homepage (My profile btn) 

Zrzut ekranu 2024-05-14 o 21 30 24

- [x] Homepage (Our team) 

Zrzut ekranu 2024-05-14 o 21 30 55

- [ ] Create / Edit profile (almost all labels) 

Zrzut ekranu 2024-05-14 o 21 31 56

- [ ] My profile 
Zrzut ekranu 2024-05-14 o 21 30 06

- [ ] We also should ensure that there are no namespaces declared as strings, I mean in every case where you have something like that:
```const t = useTranslations('Buttons')``` we should have something like that (all namespaces should be used from some enum or other predefined directory to avoid duplication):
```  const t = useTranslations(I18n.Buttons)```

src/app/[locale]/(auth)/(routes)/signin/(components)/LoginDev/LoginDev.tsx
Comment (mr-fox93) [reply src/app/[locale]/(auth)/(routes)/signin/(components)/LoginDev/LoginDev.tsx:7]: done

Sumick (approved):
Good job

meczajkowski (approved):
Great!

PR #312, opened 2024-05-16
fix: remove header from moderation view in user profile

mr-fox93:
image


Sumick (changes requested):
Im not sure if it will fix the problem as we probably can need these `children` in other place. We should remove that header from different component

mmcode9407 (changes requested):
You have to replace the `Header` with  the `UserProfileHeader` in `/moderation/page`.
And obviously, you have to leave the `children` prop in `UserProfileMain`.
Don't forget about `withBackButton` prop in `UserProfileHeader` 👍

mr-fox93:
I brought it back children in UserProfileMain.
Change Header to UserProfileHeader in page.tsx for ModerationView like said @mmcode9407

mmcode9407 (approved):
GJ ❤️

Sumick (approved):
Looks good! 🔥

PR #327, opened 2024-09-04
fix: search input is sticky now also clear btn is stick under input

mr-fox93:
image


Sumick (approved):
Looks good in code :)

PR #330, opened 2024-09-05
anchor button now ensures https

filwas:
In next.js Link element, any url not starting with the https:// protocol is being treated as an internal link. This change fixes the problem by making sure the url provided to the nextjs link element is always prepended with the protocol.

### DIFF
--- a/packages/ui-system/src/components/Button/AnchorButton.tsx
+++ b/packages/ui-system/src/components/Button/AnchorButton.tsx
@@ -2,19 +2,13 @@ import { type PropsWithChildren } from "react";
 
 import Link from "next/link";
 import styles from "./AnchorButton.module.scss";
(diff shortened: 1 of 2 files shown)


packages/ui-system/src/components/Button/AnchorButton.tsx
```diff
@@ -8,11 +8,19 @@ interface AnchorButton {
   icon?: React.ReactElement;
 }
 
+const ensureProtocol = (url: string) => {
```
Sumick, line 11: **nit:** Let's put it into `utils/routes.ts` as this probably can be reused in the future

Sumick (approved):
Great job on spotting that! Im leaving one minor comment and we are ready to go!

mr-fox93 (approved):

PR #342, opened 2024-10-21
Feat/implement currency set when creating or editing profile

filwas:


apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/CreateProfile/WorkInformation/WorkInformation.tsx
```diff
@@ -65,6 +67,9 @@ const WorkInformation = () => {
       )
     }
```
Comment (NerdbordMentor) [inline apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/CreateProfile/WorkInformation/WorkInformation.tsx:68]: **issue**: Ensure that the `onClick` handler is not called unnecessarily. Consider using a memoized callback to prevent unnecessary re-renders.

NerdbordMentor (approved):
## Overall Feedback

The implementation of the currency tab and its integration into the profile creation/editing form is well-executed. The task requirements are met, including the addition of the Currency enum in the Prisma schema and the adjustment of the hourly rate input based on the selected currency. The removal of the salary filter from the homepage is also correctly handled. The code is generally clean and follows a consistent style, making it easy to read and maintain. However, there are a few areas that could be improved for better maintainability and performance.

### Things Missing: 

- [] None, all task requirements are met.

### Things That Went Well:

- The use of enums for currency management is efficient and aligns well with the database schema.
- The conditional rendering of currency buttons is a neat solution for user interaction.
- The integration of currency selection with form validation ensures data integrity.

packages/ui-system/src/components/Button/Button.tsx
```diff
@@ -34,14 +37,17 @@ export const Button = ({
...
-    <button
-      data-testid={dataTestId}
-      className={`${buttonClasses}`}
-      disabled={disabled || loading}
-      onClick={onClick}
-      type={type}
-    >
-      {loading ? "Loading..." : children}
-    </button>
+    <label htmlFor={id}>
```
Sumick, line 40: **issue:** We break semantics by wrapping `button` into `label` in Button which is a part of ui-system

Sumick (changes requested):
Looks good but we can't wrap button in label in ui-system Button component as we break semantics this way

Sumick (approved):
Looks good!

PR #366, opened 2024-12-01
feat: Auth v3 - Enhanced Authentication System & LinkedIn Provider!!!

meczajkowski:
Finally! 

# feat: Auth v3 - Enhanced Authentication System & LinkedIn Provider!!!
### Overview 
This PR significantly upgrades our authentication system by introducing LinkedIn as a provider for Specialists and Hunters. Users can now register and log in using LinkedIn, GitHub, or both. This update also optimizes JWT handling, enhances access control, and adds support for custom slugs in user profiles.

Additionally, database calls within JWT and session callbacks have been minimized. The JWT callback now only queries the database during sign-in, reducing unnecessary calls during the session lifecycle.

### Key Changes

- _**LinkedIn Integration for Specialists and Hunters:**_ Specialists and Hunters can now sign up and log in via LinkedIn. This provides more flexibility, especially for users without GitHub accounts.

- _**Email Provider for Hunters:**_ Email authentication is currently available exclusively for Hunters. However, extending it to Specialists is straightforward

- _**Custom Slug Creation for Profiles:**_ Since LinkedIn doesn’t provide a username, users can now create custom slugs during profile creation. For GitHub users, the slug field in the profile creation form is pre-filled with their GitHub usernamebut remains editable.

- **_Account Merging:_** If a user registers with the same email through both LinkedIn and GitHub, their accounts will be seamlessly merged.

- **_Efficient JWT Handling:_** Database calls in JWT and session callbacks have been minimized. The JWT callback now fetches user data from the database only during sign-in, eliminating hundreds of redundant calls during a session lifecycle.

- **_Discord Notifications:_** 

Sign-up notifications now include the authentication provider (LinkedIn or GitHub), providing better context about the user’s origin. Notifications are also sent if a user attempts to add an unauthorized role, including their identifier for easier tracking.

### Technical Details

- Added `slug `column to `Profile `model in the Prisma schema to store custom slugs.
- Refactored role-based access logic to separate and streamline access controls for Hunters and Specialists, enhancing maintainability and security.

### TODO

- [x] Custom Slug Validation: Implement a slug validation mechanism to prevent duplicate or inappropriate slugs.
- [x] GitHub Username Updates: Ensure that GitHub usernames are updated during sign-in. Currently, they are only set during sign-up and don't reflect changes made on GitHub.
- [x] Specialist Role Validation for Email Provider: Add checks to prevent Specialists from signing in via the email provider, avoiding incorrect welcome emails intended for Hunters.

PR #369, opened 2024-12-03
feat: add optional slug column to profile model (Auth v3 step 1)

meczajkowski:
## Summary
This change is a preparatory step for merging the Auth v3 PR (#366) and should be merged first to ensure a smooth transition.

#### New Column Introduction:
Adds a slug column to the Profile model, initially set as optional. This allows time to generate slugs for existing users without disrupting the current workflow.

#### Slug Population Plan:
After deploying this PR to production:

1. Run a query to populate all slugs using GitHub usernames from the githubDetails model.
2. Merge PR #366, which makes the slug field required.

**Important**: These steps (this PR, the query, and #366) should be executed sequentially, with minimal time between each, to prevent profiles from being created without slugs during the transition.

## Testing
This change has been tested and ensures that:

- Existing user profiles load and can be edited as expected.
- New users can create, edit, and interact with profiles seamlessly.
- Existing users can log in, log out, and create profiles as before.
- New signups and logins function correctly.

**Note**: This update does not impact the current application flow.

### DIFF
--- a/apps/platform-app/prisma/migrations/20241203132444_add_optional_slug_column_to_profile_table/migration.sql
+++ b/apps/platform-app/prisma/migrations/20241203132444_add_optional_slug_column_to_profile_table/migration.sql
@@ -1,0 +1,11 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[slug]` on the table `Profile` will be added. If there are existing duplicate values, this will fail.
+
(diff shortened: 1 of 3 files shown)


transcendence12 (approved):

Sumick (approved):
Let's gooo

PR #371, opened 2024-12-09
Feat/implement resume upload

transcendence12:
**Fix issues:**

- modify copy on Upload cv button,
- handle unnecessary error during upload,
- make possible change cv file,
- make possible download cv


apps/platform-app/src/app/[locale]/(profile)/(components)/CVuploader/CvUploaderForm.tsx
```diff
@@ -4,25 +4,58 @@ import { useUploadContext } from '@/contexts/UploadContext'
...
+    url: string
+  } | null>(null)
 
   const { cvUploadError, onSetCvUploadError, onSetCvFormData } =
     useUploadContext()
-  const [selectedFile, setSelectedFile] = useState<File | null>(null)
+
   const [errorMsg, setErrorMsg] = useState<string | null>(null)
 
+  useEffect(() => {
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(components)/CVuploader/CvUploaderForm.tsx:32]: **nit:** Do we really need to fetch this URL here if its a part of ProfileModel which is fetched to form state? I think we can extract this url from form state

apps/platform-app/src/app/[locale]/(profile)/_actions/mutations/fetchCVurl.ts
```diff
@@ -0,0 +1,11 @@
+'use server'
+
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/_actions/mutations/fetchCVurl.ts:2]: **nit:** This is query not mutation. Btw, I think we can remove this anyway, as CV URL is included in ProfileModel that is fetched when initializing form. Such url should be available in form values object.

Sumick (changes requested):
Let's ensure that we really need this fetchCVUrl logic, because when we enter edit profile form, we already have CV URL included in ProfileModel object, so basically we could just pass it as prop or extract from form state.

apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/CreateProfile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -65,7 +69,7 @@ const PersonalInfo = () => {
           />
         </div>
         <UserPhotoUploader />
-        <CVUploader />
+        <CVUploader initialCvUrl={initialCvUrl} />
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/CreateProfile/PersonalInfo/PersonalInfo.tsx:72]: **nit:** no need to pass prop that deep, we can use:

```ts
 const { values, handleChange, errors, touched, handleBlur } =
    useFormikContext<CreateProfileFormValues>()
```

apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/EditProfileForm/EditProfileForm.tsx
```diff
@@ -102,7 +102,7 @@ const EditProfileForm = ({ profile }: { profile: ProfileModel }) => {
       avatarUrl: session?.user?.image || null,
       linkedIn: values.linkedin,
       bio: values.bio,
-      cvUrl: values.cvUrl,
+      cvUrl: values.cvUrl ?? profile.cvUrl,
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(routes)/my-profile/(components)/EditProfileForm/EditProfileForm.tsx:105]: **nit:** I think it should not be neccesary here, as values.cvUrl should be already set when initalizing from data

apps/platform-app/src/app/[locale]/(profile)/(components)/CVuploader/CvUploaderForm.tsx
```diff
@@ -8,21 +8,37 @@ import { useTranslations } from 'next-intl'
...
 
-export function CVuploaderForm() {
+interface CvUploaderFormProps {
+  initialCvUrl: string | null
+}
+
+export function CVuploaderForm({ initialCvUrl }: CvUploaderFormProps) {
   const t = useTranslations(I18nNamespaces.Buttons)
+  const tt = useTranslations(I18nNamespaces.PersonalInfo)
   const [isUploading, setIsUploading] = useState(false)
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(components)/CVuploader/CvUploaderForm.tsx:19]: **nit:** we can extract form value by using useFormixContext, and extract inital url from there without need of passing down this prop that many levels:

```ts
 const { values } =
    useFormikContext<CreateProfileFormValues>()
```

Sumick (changes requested):
We should not need such prop drilling (passing prop many levels down) because we can use in CvUploaderForm:

```ts
 const { values, handleChange, errors, touched, handleBlur } =
    useFormikContext<CreateProfileFormValues>()
```

and our prop will be:

```ts
const cvUrl = values.cvFormUrl
```

Sumick (approved):
Looks good now!

filwas (approved):

meczajkowski (approved):

PR #382, opened 2025-02-09
Feat/implement frontend for /jobs/add, /jobs/:id and /jobs/:id/edit pages

Wiecek-K:


Sumick (approved):
Few things found, but overall looks really good! Let's leave AI related things to refactor for later when we will implement that specifically

PR #386, opened 2025-02-28
RELEASE: Dark / light mode + bug fixes

Sumick:


Wiecek-K (commented):
A few suggestions in code, and:
- I think, a screen shoot with the same theme as a page theme will fit more here
image
- SOMEBODY KIDNAPPED OUR TEAM!!!" 😱, and the header is in the same color as the other lines
image
- When I select the 'frotnend-dev' tag in filters, the tag does not get an additional color. Theme doesn't matter here.
image
image
- badges on dev profile (on my-profile it's ok)
image

hansac91 (approved):

Wiecek-K (approved):

PR #391, opened 2025-03-03
Feat/client landing setup

AniaPiWo:
I have completed the initial setup for the hunter's landing page route. Each section is a separate component, and all components already have their corresponding stylesheet connected.


Sumick (approved):
🎸

PR #396, opened 2025-03-05
Feat/hunter faq section

AniaPiWo:
the section includes an interactive accordion with  animation, a "Write to us" CTA button and translations


apps/platform-app/src/components/hunter-landing/HunterFAQ/HunterFAQ.tsx
```diff
@@ -1,7 +1,96 @@
...
+type AccordionItem = {
+  question: string
+  answer: string
+}
+
+type AccordionProps = {
+  items: AccordionItem[]
+}
+
+const Accordion = ({ items }: AccordionProps) => {
```
Sumick, line 62: **nit:** This should be in separate file, src/components/hunter-landing/Accordion

Sumick, line 62: **issue:** Naming is incorrect, as you made a list, so it should be AccordionGroup. We should split it into Accordion, and AccordionGroup as currently it's not possible to use single accordion without section wrapping

Sumick (commented):
Mostly looks good, but we need to split these components nicely before merging

Sumick (approved):

PR #400, opened 2025-03-13
Feat/ header

AniaPiWo:
The header component for  desktop and mobile, featuring a custom-styled LocaleSwitcherSelect for the Hunter Landing page. A new variant="hunterLanding" prop was introduced, applying specific styles while keeping the default ones unchanged.


apps/platform-app/src/components/hunter-landing/HunterHeader/HunterHeader.tsx
```diff
@@ -1,7 +1,67 @@
...
+    } else {
+      setIsMobileMenuOpen(true)
+      setTimeout(() => {
+        setMenuVisible(true)
+      }, 10)
+    }
+  }
+
+  return (
+    <section id="HunterHeader" className={styles.wrapper}>
```
Wiecek-K, line 37: Id HunterHeader in section tag - it is a little confusing
Btw is better to put that into <Container/> component to avoid problems with display on huge screens
image

apps/platform-app/src/components/hunter-landing/HunterHeader/HunterHeader.tsx
```diff
@@ -1,7 +1,67 @@
...
+      <div className={styles.menuDesktop}>
+        <LocaleSwitcher variant="hunterLanding" />
+
+        <Button variant="allpurple" onClick={handleAddJob}>
+          {t('addJob')}
+        </Button>
+      </div>
+      <div className={styles.menuMobile}>
+        <button className={styles.mobilebutton} onClick={toggleMobileMenu}>
+          {isMobileMenuOpen ? 'X' : 'Menu'}
```
Wiecek-K, line 48: You replace "Menu" with "X" but don't change the button size - Consider using "Close" instead of "X"

apps/platform-app/src/app/[locale]/page.tsx
```diff
@@ -17,16 +17,18 @@ export default async function HunterLandingPage() {
...
-        <HunterRiskReducers />
-        <HunterHowItWorks />
-        <HunterWhyWorth />
-        <HunterReviews />
-        <HunterTeam />
-        <HunterCTA />
-        <HunterFAQ />
-        <HunterFooter />
+        <div style={{ display: 'flex', flexDirection: 'column', gap: '180px' }}>
+          <HunterHeader />
```
Comment (Wiecek-K) [inline apps/platform-app/src/app/[locale]/page.tsx:21]: Move <hunterHeader/> and <hunterFooter> outside the <main/> tag to improve semantics

Wiecek-K (changes requested):
The locale switcher doesn't show options on mobile view - probably they are hidden under the menu wrapper
image

apps/platform-app/src/app/[locale]/(profile)/(components)/LocaleSwitcher/LocaleSwitcher.tsx
```diff
@@ -3,9 +3,15 @@ import { I18nNamespaces } from '@/i18n/request'
 import { useLocale, useTranslations } from 'next-intl'
 import { LocaleSwitcherSelect } from './LocaleSwitcherSelect'
 
-export const LocaleSwitcher = () => {
+export const LocaleSwitcher = ({ variant }: { variant?: 'hunterLanding' }) => {
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(components)/LocaleSwitcher/LocaleSwitcher.tsx:6]: let's move variants to enum, so we can export them across project

apps/platform-app/src/app/[locale]/(profile)/(components)/LocaleSwitcher/LocaleSwitcherSelect.tsx
```diff
@@ -55,31 +57,24 @@ export const LocaleSwitcherSelect = ({
     })
   }
 
+  const containerClassName = cx('container', {
+    open: isOpen,
+    hunterLanding: variant === 'hunterLanding',
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(profile)/(components)/LocaleSwitcher/LocaleSwitcherSelect.tsx:62]: then let's use enum here to not maintain strings

apps/platform-app/src/app/[locale]/page.tsx
```diff
@@ -17,16 +17,18 @@ export default async function HunterLandingPage() {
...
-        <HunterHero />
-        <HunterRiskReducers />
-        <HunterHowItWorks />
-        <HunterWhyWorth />
-        <HunterReviews />
-        <HunterTeam />
-        <HunterCTA />
-        <HunterFAQ />
-        <HunterFooter />
+        <div style={{ display: 'flex', flexDirection: 'column', gap: '180px' }}>
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/page.tsx:20]: we should make a class for that, we should not style inline like this

apps/platform-app/src/components/hunter-landing/HunterHeader/HunterHeader.tsx
```diff
@@ -1,7 +1,67 @@
...
-export const HunterHeader = (props: Props) => {
-  return <div className={styles.wrapper}> HunterHeader</div>
+  const handleAddJob = () => {
+    router.push(AppRoutes.postJob)
+  }
+
+  const toggleMobileMenu = () => {
+    if (isMobileMenuOpen) {
+      setMenuVisible(false)
+      setTimeout(() => {
```
Sumick, line 25: why timeout? we should handle it via pure CSS

Sumick (commented):
Code looks mostly good - I haven't tested UI yet, but I know Ajmag did and I trust you on the accuracy against Figma designs.

PR #410, opened 2025-03-29
feat: create  LoginToApplyModal

Wiecek-K:
The task located on the Kanban board with the title "Szlifowanie procesu aplikacji do zlecenia"


Sumick (changes requested):
I found 2 issues while testing:

- [ ] I can't upload CV or display it, I don't know. Looks like something happened to CV uploader
- [ ] There is a problem with our TextEditor. Block type cannot be changed, and it looks like UI crashed. Maybe we should upgrade editor package version?
Zrzut ekranu 2025-03-30 o 16 08 59

PR #416, opened 2025-04-07
Job verification and candidate matching for job

Wiecek-K:
task on nerdbort: "Szlifowanie weryfikacji zlecenia"


apps/platform-app/src/app/[locale]/(jobs)/_workflows/matchJobWithProfiles.ts
```diff
@@ -77,37 +100,32 @@ export async function matchJobWithProfiles(
...
       },
     ]
 
-    // Generate the response with JSON format
-    const jsonResponse = await groqService.generateResponse(messages, {
-      json: true,
+    const matchesResult = await openaiService.generateJson<{
+      matches: MatchResult[]
+    }>(messages, {
+      model: 'gpt-4-turbo',
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(jobs)/_workflows/matchJobWithProfiles.ts:121]: we should use `gpt-4o` here

apps/platform-app/src/app/[locale]/(jobs)/_workflows/notifyMatchedProfiles.ts
```diff
@@ -13,14 +13,10 @@ export async function notifyMatchedProfiles(
   sentCount: number
 }> {
   try {
-    // Filter matches with score above threshold (e.g., 70)
-    const goodMatches = matchingResults.filter(
-      (match) => match.matchScore >= 70,
-    )
     const approvedProfiles = await findAllApprovedProfiles()
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(jobs)/_workflows/notifyMatchedProfiles.ts:16]: if we return from LLM array of objects with `profileId` included, we could skip all this finding and sorting method as we already have profile IDs and we can do one simple find call and remove this redundant logic

apps/platform-app/src/app/[locale]/(jobs)/_workflows/matchJobWithProfiles.ts
```diff
@@ -77,37 +100,32 @@ export async function matchJobWithProfiles(
...
       },
       {
-        role: 'user',
+        role: 'user' as const,
         content: JSON.stringify({
           job: jobData,
-          profiles: profilesData,
         }),
       },
     ]
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(jobs)/_workflows/matchJobWithProfiles.ts:117]: im not sure if we should place it here or in `notifyMatchedProfiles` but we need to update relations in DB, so we can track to whom we did sent jobs to. then we will track who responds to it, and who is inactive. also, we will need a mechanism for sending another 10 people if no-one responds in eg. 12 hours

Sumick (changes requested):
Mostly looks really good! I found 3 additional very important things that we should also implement in that task scope

apps/platform-app/src/backend/jobCandidate/jobCandidate.service.ts
Sumick, line 1: we should keep our file naming consistent in kebab case, so `job-candidate`,

Sumick (approved):
Looks good! Im leaving one comment + we can move this saving job candidates logic to matchmaking logic before merging

PR #421, opened 2025-04-14
Feat/save preferred language in database

Wiecek-K:


apps/platform-app/prisma/schema.prisma
```diff
@@ -16,15 +16,16 @@ model User {
   nerdbordUserId       String?          @unique
   emailVerified        DateTime?
   createdAt            DateTime         @default(now())
+  preferredLanguage    String           @default("")
```
Sumick, line 19: this should be default value set her form `en` -  and also languages should be predeclared enums so we dont do mistakes as plcing some random string here

apps/platform-app/src/app/[locale]/(auth)/_models/User.model.ts
```diff
@@ -6,6 +6,7 @@ export interface UserModel extends User {
   githubUsername: string | null
   profileId: Profile['id'] | null
   profileSlug: Profile['slug'] | null
+  preferredLanguage: string
```
Comment (Sumick) [inline apps/platform-app/src/app/[locale]/(auth)/_models/User.model.ts:9]: we should type available languages as its too error prone

Sumick (changes requested):
We need some improvements and cleanup, however good job on solving that problem!

Sumick (approved):
Looks good now, good job 🫡

